PR #47, opened 2013-04-17
Make fallback optional

kategray:
Make falling back to the default locale optional, so that translate () can be used to force creation, even when a translation in the default locale is already set.


taylankasap:
Why is this issue still open?

PR #75, opened 2013-11-06
Add possibility to pass extra parameters in getTree

raziel057:
It would be usefull to be able to pass parameters that can be retrived in the method addFlatTreeConditions.

For example if you need to add conditions depending on the user rights or others...


Einenlum:
PR was rebased here: https://github.com/KnpLabs/DoctrineBehaviors/pull/358

Closing this one.

PR #82, opened 2013-12-06
[Translatable] Fix property access on twig

asosso:
This should fix this warning. 

An exception has been thrown during the rendering of a template ("Warning: call_user_func_array() expects parameter 1 to be a valid callback, class 'MyClassTranslation' does not have a method 'name' in ../vendor/knplabs/doctrine-behaviors/src/Knp/DoctrineBehaviors/Model/Translatable/TranslatableMethods.php line 140") in MyBundle:Default:index.html.twig at line 1.

Twig calls 'name' instead 'getName'

### DIFF
--- a/src/Knp/DoctrineBehaviors/Model/Translatable/TranslatableMethods.php
+++ b/src/Knp/DoctrineBehaviors/Model/Translatable/TranslatableMethods.php
@@ -134,6 +134,9 @@ trait TranslatableMethods
 
     protected function proxyCurrentLocaleTranslation($method, array $arguments = [])
     {
+        if (!method_exists(self::getTranslationEntityClass(), $method))
+            $method = 'get'. ucfirst($method);
+
         return call_user_func_array(
             [$this->translate($this->getCurrentLocale()), $method],
             $arguments


docteurklein:
I changed my mind. Such a small change could maybe be merged. If you rebase I will merge it. Sry :)

lemoinem:
Hum...

Stupid question, why not use Symfony's PropertyAccess?

It would be even cleaner...

docteurklein:
Not sure I'd want PropertyAccess in my domain model.

lemoinem:
What would be the issue? It does give access to anything you wouldn't already have anyway (only uses public methods or properties). And it literally is its to abstract away this kind of plumbing. 

It also addresses the "No Twig/library specific code in Translation" concern. PropertyAccessor goal is to provide  a usage agnostic way to access another object public properties.

Plus if it's hidden in a Trait, it doesn't seem much worse than already implementing the magic `__call` (in a Trait). Since we're using magic already, we might as well use some robust and future friendly one). It supports both Twig Stlye `$car->name()` and PHP Style `$car->getName()`.

I agree it doesn't feel clean for regular use, but in a Decorator/Proxy perspective, it seems the way to go.

The get-only version is a one-liner simple:

```php
public function __call($method, $arguments)
{
     return PropertyAccess::createPropertyAccessor()->getValue($this->translate(), $method);
}
```

The get/set version isn't much more complex (asserts could be replaced by `if(!CONDITION) return;`):

```php
public function __call($method, $arguments)
{
    $translation = $this->translate();
    $propertyAccessor = PropertyAccess::createPropertyAccessor();

    if (in_array(substr($method, 0, 3), ['set', 'add']) || 'remove' === substr($method, 0, 6))
    {
       assert('1 === count($arguments)');
       $propertyAccessor->setValue($translation, $method, reset($arguments));

       return $this;
   }

    assert('0 === count($arguments)');
    return $propertyAccessor->getValue($translation, $method);
}
```

TomasVotruba:
Replaced by https://github.com/KnpLabs/DoctrineBehaviors/pull/440

PR #94, opened 2014-02-15
Extract slug generator to separate method

enl:
The main reason for changing this is that all standard php methods of transliaterationg are unstable (like iconv) or unusable for russian (Transliterator class). So that I managed to extract slug generation to separate method which can be redefined without need to copypaste.

In addition, I created Utils class with static function which applies traditional russian way of transliteration.


MAXakaWIZARD:
@enl I think this issue has been fixed in #162

PR #95, opened 2014-02-22
getField

asosso:
Use the getter instead of calling directly the variable.
This ensures compatibility with Translatable


Einenlum:
@docteurklein What do you think today of this?

docteurklein:
still not really interesting to me

docteurklein:
just override the method if you don't like the default impl.

Einenlum:
@docteurklein Kay, thanks for your answer. I close this PR then.

PR #105, opened 2014-04-01
Analyze traits recursive

bvleur:
Detect a trait when using them via an intermediate trait.


docteurklein:
@bvleur thanks, but I think we'll switch to real interfaces to detect entities, which makes the class analyzer obsolete.

PR #112, opened 2014-05-26
Adds a parameter for the fetch method used by doctrine for the translations

bobvandevijver:
When using doctrine in lazy mode and you need to display a lot record with translations on the same page, every translate call will result in a extra query to fetch the translation that is needed. 

When the fetch mode is set to EAGER, this problem will not occur, but it might have a negative impact on the performance when it is not needed. Therefore the default is still LAZY, and nothing changes in the behavior with this PR.


ruscon:
i'm trying to use:

```
knp.doctrine_behaviors.translatable_subscriber.translatable_fetch_method: EAGER
knp.doctrine_behaviors.translatable_subscriber.translation_fetch_method: LAZY
```
it's not work

if i'm trying to use:
```
knp.doctrine_behaviors.translatable_subscriber.translatable_fetch_method: LAZY
knp.doctrine_behaviors.translatable_subscriber.translation_fetch_method: EAGER
```
it's wokring, but not need for me :)

PR #116, opened 2014-06-11
[Loggable] Parametrized loggable trait

gaydarov:
The loggable trait is also now adjusted, so that it is possible to use own loggable trait. Tests have been also updated.


TomasVotruba:
Alread covered with `Knp\DoctrineBehaviors\Contract\Entity\LoggableInterface`

PR #159, opened 2014-10-22
[Translatable] Restore BC-break in Translation Traits

bobvandevijver:
I think this should be enough to restore the BC-break introduced in
75e1187fbf4a116901491c5d31165b8824d589eb.


TomasVotruba:
Closing as this issues is almost 6 years old

PR #161, opened 2014-10-26
Try to use getters when getting non-existent field values for Sluggable

MAXakaWIZARD:
Use getters when getting field values for Sluggable, if property does not exist (solves problem with accessing non-existent field, like in translations)

### DIFF
--- a/src/Model/Sluggable/SluggableMethods.php
+++ b/src/Model/Sluggable/SluggableMethods.php
@@ -74,9 +74,17 @@ trait SluggableMethods
             $usableValues = [];
 
             foreach ($fields as $field) {
-                // Too bad empty is a language construct...otherwise we could use the return value in a write context :)
-                $methodName = 'get' . ucfirst($field);
-                $val = $this->{$methodName}();
+                if (property_exists($this, $field)) {
+                    $val = $this->{$field};
+                } else {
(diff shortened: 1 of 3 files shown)


EmmanuelVella:
:+1: I need this feature too, a merge would be great !

src/Model/Sluggable/SluggableMethods.php
```diff
@@ -74,8 +74,17 @@ public function generateSlug()
...
+                if (property_exists($this, $field)) {
+                    $val = $this->{$field};
+                } else {
+                    $methodName = 'get' . ucfirst($field);
+                    if (method_exists($this, $methodName)) {
+                        $val = $this->{$methodName}();
+                    } else {
+                        $val = null;
+                    }
+                }
```
EmmanuelVella: Maybe something like this would be more clear : 

```
$methodName = 'get'.ucfirst($field);

if (method_exists($this, $methodName)) {
    $val = $this->{$methodName}();
} elseif (property_exists($this, $field)) {
    $val = $this->{$field};
} else {
    continue;
}
```

MAXakaWIZARD:
@docteurklein please tell us your opinion, this PR hangs here for 5 months already.

docteurklein:
LGTM! sorry for the delay, i'm quite busy these days.

PR #162, opened 2014-10-27
Slug generation from cyrillic strings

MAXakaWIZARD:
Replace iconv usage with behat/transliterator, add some test cases.
Slug generation from cyrillic string now works.


MAXakaWIZARD:
@docteurklein please review this

PR #167, opened 2014-12-13
Allow locale as an association entity

burci:
This allows to have ManyToOne association as location

```
/**
 * @var \Acme\Locale $locale
 *
 * @ORM\ManyToOne(targetEntity="Locale")
 * @ORM\JoinColumn(name="locale", referencedColumnName="id", nullable=false)
 */
protected $locale;
```

### DIFF
--- a/src/Model/Translatable/TranslatableMethods.php
+++ b/src/Model/Translatable/TranslatableMethods.php
@@ -49,7 +49,7 @@ trait TranslatableMethods
      */
     public function addTranslation($translation)
     {
-        $this->getTranslations()->set($translation->getLocale(), $translation);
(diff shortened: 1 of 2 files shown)


akovalyov:
@docteurklein what do you think with this one?
Looks good for me, I can add some docs for it or @burci could if he has time and willing.

akovalyov:
Also there is should be mentioned that ```Locale``` custom class must implement ```__toString()``` method, otherwise it could not be casted on https://github.com/KnpLabs/DoctrineBehaviors/pull/167/files#diff-70fa0cb795057a4d16c686305571a662R52 and https://github.com/KnpLabs/DoctrineBehaviors/pull/167/files#diff-70fa0cb795057a4d16c686305571a662R102

Issue #172, opened 2015-01-03
[Translatable] Feature: ability to choose the class translation name


asprega:
What about adding a custom annotation (or even a dead simple static method) to translatable classes to override the default `<classname>Translation` class used to establish the mapping?

How it is now:
Acme\Model\Category <=> Acme\Model\CategoryTranslation

How it could be with the proposed feature:
Acme\Model\Category <=> Acme\Model\Translation\CategoryTranslation

The reason why I'm proposing this is that I find very useful to separate entities and translations in different namespaces: this allows for less pollution and better organization when you have many translatable entities.

If you guys think this may be a nice addition, I can provide a pull request.

docteurklein:
I thought it was configurable through a static method, but it appears it isn't https://github.com/KnpLabs/DoctrineBehaviors/blob/master/src/ORM/Translatable/TranslatableSubscriber.php#L206.

Feel free to PR something.

I'm not sure i'm willing to introduce annotations (even tho it would be a great way to configure it :) )
It's just that until now, we didn't need annotations.

Anyway, let's try something.

asprega:
I agree with you, an annotation would seem overkill... But you already require doctrine/common which in turns requires doctrine/annotations, so no additional dependencies would be (practically) needed. There may be other use cases for custom annotations in other behaviors (just guessing, didn't look at them yet).

Alternatively, would you prefer using an interface to dictate which static method should be implemented? Or just duck typing?

docteurklein:
a trait method might be enough.

nick4fake:
https://github.com/KnpLabs/DoctrineBehaviors/pull/166

asprega:
The trait actually already has a method to return the translation class (https://github.com/KnpLabs/DoctrineBehaviors/blob/master/src/Model/Translatable/TranslatableMethods.php#L152). There are 2 problems:
1. It isn't currently used by the listener which establishes the mapping (only by the trait itself)
2. AFAIK it is not possible to override trait methods without triggering an E_NOTICE which, depending on your error reporting settings, might cause fatals (in my case, it does with production settings).

Alternatives would seem to be:
- using an annotation
- implementing an additional interface (only specific for overriding the translation name)
- quick and dirty duck typing without overriding the trait method.

Please let me know if I'm missing something.

docteurklein:
@asprega woah! thanks for such a detailled answer!

I forgot about this method. I guess we should use it in the listener then.
[...]

PR #174, opened 2015-01-15
Issue #172: ability to choose the class translation name

asprega:
- As discussed, I enforced the use of the trait methods in the translatable listener
- I created the inverse method to retrieve the in the translation entity
- I created 2 more fixture entities (I'm not 100% happy with their naming but I couldn't come up with a better one). I put the translation into a nested namespace to "break" the default rule of the "Translation" suffix.
- I added a test for verifying the oneToMany relationship also for the "customized" case. Since the 2 tests were equal, I factored its logic into a private assertion method in the test case class which now takes as parameters the 2 classes to verify the association for.
- I updated the relevant section in the readme

Hope this is ok, let me know if I need to change anything :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,11 @@ You now have a working `Category` that behaves like:
 ### translatable:
 
 If you're working on a `Category` entity, the `Translatable` behavior expects a
-**CategoryTranslation** entity.  This naming convention avoids you to handle manually entity associations.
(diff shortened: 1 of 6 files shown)


docteurklein:
seems good to me!

asprega:
Hi :) any news about when this will be merged?

docteurklein:
thanks!

asprega:
You're welcome :) Thanks for merging it.

Issue #100, opened 2014-03-10
[Translatable] Example how print translatable values in twig (knp translatable behaviour)


kvaje:
Hi,

Is any example available how print translatable field values in twig?

grekpg:
i think about this way

`/**
     * @return string
     */
    public function getContent()
    {
        if($this->translate()->getLocale()!=AppBundle::DEFAULT_LOCALE)
            return $this->translate()->getContent();

        return $this->content;
    }
`

in this way default locale i store in original entity and user next db query only for translations. 

but for simple 

{{category.getName}}

its enough in main entity

  `
public function getContent(){
return $this->translate()->getContent();
}
`

and we have current lang version and ide helper for getContent method

patie:
any news on this?

jkhamler:
I'm struggling with this myself - using Symfony 4 and an absolute vanilla use case. Product and ProductTranslation. Adding the following to my Product class avoids a fatal error but it's returning blank strings.

``` 
public function __call($method, $arguments)
    {
        return $this->proxyCurrentLocaleTranslation($method, $arguments);
    }
```

TomasVotruba:
Resolved as answered

bor-yo:
This is how I solved it, hope it helps to someone

`class Category implements TranslatableInterface
{
    use TranslatableTrait;
    /**
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     */
    private $id;

    private $name;

    public function getId(): ?int
    {
        return $this->id;
    }

    public function getName()
    {
        if(!$this->translations[$this->getCurrentLocale()]) {
            return null;
        }
        return $this->translations[$this->getCurrentLocale()]->getName();
    }
}`

bor-yo:
This is how I solved it
`class Category implements TranslatableInterface
{
    use TranslatableTrait;
    /**
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     */
    private $id;
    
    private $name;

    
    public function getName()
    {
        if(!$this->translations[$this->getCurrentLocale()]) {
            return;
        }
        $this->name = $this->translations[$this->getCurrentLocale()]->getName();
        return $this->name;
    }
}`

PR #184, opened 2015-02-24
[Translatable] Compatible with class property and method

markitosgv:
In some cases like twig call like {{ post.title }} doen´t work cause first search is property, if we force to getter method always works.

### DIFF
--- a/src/Model/Translatable/TranslatableMethods.php
+++ b/src/Model/Translatable/TranslatableMethods.php
@@ -138,6 +138,8 @@ trait TranslatableMethods
 
     protected function proxyCurrentLocaleTranslation($method, array $arguments = [])
     {
+        $method = ('get' === substr($method, 0, 3)) ? $method : 'get'. ucfirst($method);
+
         return call_user_func_array(
             [$this->translate($this->getCurrentLocale()), $method],
             $arguments


francescolaffi:
you should probably check the discussions on https://github.com/KnpLabs/DoctrineBehaviors/pull/82 and https://github.com/KnpLabs/DoctrineBehaviors/issues/100

I agree with the opinion that its not a concern for this library to support twig directly

Imho a viable solution is to just explicitly define the getters on the translatable entity and make them act as a proxy, they'll work in twig, they'll work on static code analysis and also on the smart features of IDEs, no magic involved. Especially if you consider the translatable entity to be acting as root for the translatable+translations aggregate the proxied getters should be probably part of the translatable entity interface

docteurklein:
What about adding a `__get` proxy ?

francescolaffi:
@docteurklein for __get to work in twig there must also be __isset returning true for overloaded properties

anyway I would prefer the Translatable trait not to do overloading by default, maybe there could be other proxy methods to help defining __get and __isset, or another trait with them defined

TomasVotruba:
Replaced by https://github.com/KnpLabs/DoctrineBehaviors/pull/440

PR #186, opened 2015-02-27
Use symfony property access component to get sluggable fields values

EmmanuelVella:
This PR replaces the custom property/method access with the [symfony/property-access component](http://symfony.com/doc/current/components/property_access/introduction.html) which is much more powerful. Ex : 

```
public function getSluggableFields()
{
    return ['category.name', 'title'];
}
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,8 +21,9 @@
         }
     ],
     "require": {
-        "php":              ">=5.4",
-        "doctrine/common":  ">=2.2"
+        "php":                     ">=5.4",
+        "doctrine/common":         ">=2.2",
+        "symfony/property-access": ">=2.3"
     },
     "require-dev": {
         "ext-pdo_sqlite":                   "*",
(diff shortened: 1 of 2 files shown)


EmmanuelVella:
Coule someone please have a look at it ? Thank you ;)

docteurklein:
is it a good idea to tie the model to this component?
Maybe we should build the slug directly in the model,
and let the listener do only the transliteration?

docteurklein:
this way no need of  property path, only php.

docteurklein:
and encapsulation is preserved

docteurklein:
or we keep it like this, but move the property access in the listener (i prefer to « build the slug directly in the model»)

EmmanuelVella:
Thank you for your feedback !

Yes both ideas are good : 
- Create a ``SluggableMethods::getSluggableString`` with a BC (fallback on ``getSluggableFields``)
- Move the property access to the listener : I think this would cause a BC break as the ``getSluggableFields``, etc. will need to be public.

Einenlum:
I think this PR can be closed then. Feel free to answer if you think this should be reopen.

PR #187, opened 2015-02-27
Call generateSlug from SluggableSubscriber

EmmanuelVella:
**Use case** : I need to create a slug from a field which is updated in a entity listener.

**Problem** : The ``generateSlug`` method is called before any other listener/subscriber, because it is registered as lifecycle events which are prioritary (no matter the "priority" attribute).

**Solution** : Call the ``generateSlug`` method from the ``SluggableSubscriber``.


docteurklein:
thanks!

PR #189, opened 2015-03-09
added customizable tree identifier

digitalkaoz:
with this patch it is possible to another key than the `ID` for generating the materialized path. (i wanted to use the slug field instead)

you can override `getTreeNodeField` for that purpose:

```php
    public function getNodeId()
    {
        if (!$this->getSlug()) {
            $this->generateSlug();
        }

        return $this->getSlug();
    }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,8 @@ You now have a working `Category` that behaves like:
 
 ```
 
+> it is possible to use another identifier than `id`, simply override `getTreeNodeField` and return your custom identifier (works great in combination with `Sluggable`)
+
 <a name="translatable" id="translatable"></a>
 ### translatable:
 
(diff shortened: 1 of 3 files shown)


docteurklein:
good idea, but i'm not sure about the method name. what about `getNodeId` ?

digitalkaoz:
you are free to wish a name :)

docteurklein:
how fast :)

digitalkaoz:
im a bit unsure about the repository (its working with ids), but its works perfectly here...maybe i just didnt covered side-effects?

digitalkaoz:
a i would love to see it quickly in a stable tag :) so i can remove my branch ;)

docteurklein:
the repository works with the resulting materialized path, so it has no knowledge of `id`. 
it will be a major realease since it's a BC break. Thanks!

digitalkaoz:
mh i dont see a BC...the default behavior falls back to `getId`

or am i missing something?

docteurklein:
hooo, right :) nevermind

digitalkaoz:
what about merging then? @docteurklein

docteurklein:
was waiting for travis and forgot :) thanks!

digitalkaoz:
merci bien :)

Issue #182, opened 2015-02-23
Provide default locale for Translatable


charlie-wasp:
Hello! I am searching for the way to set provide default locale for `translate` method. Unfortunately, I found out this in `TranslatableMethods`:

```
    public function getDefaultLocale()
    {
        return 'en';
    }
```

So, I guess, there is no way to override it? :smile:

kuczek:
@charlieWasp If you are searching for default locale you probably should $translator->getFallbackLocales()[0] (with isset probably), and it also can be done like that here (i think it should) @docteurklein?

kuczek:
Probably we should think how to use it with different language providers, in symfony it will be defaultFallback, and all fallbacks should be supported...

markitosgv:
Im using in a Symfony app, and i think the best way is modify this method to read a configuration key to set this locale.

But its a trait, and i don´t know how we can configure this value, maybe a setter method?

kuczek:
@markitosgv The bigest problem is, that this is not bundle, and not only symfony, so we need to prepare universal mechanism.

charlieWasp:
@kuczek can we move default locale to parameter in `orm-services.yml`? So clients will be able to override it. Or maybe we can use callable class for it

markitosgv:
@kuczek i know, maybe @charlieWasp proposal to move parameter to .yml is a good solution

kuczek:
@markitosgv @charlieWasp But what about non Symfony usage? I agree that we should somehow make it settable, but maybe we should wrote/edit some listener that will inject this from somewhere. For symfony it will be the best solution, for the rest it will be possible to do that, and default value can stay as [en].

Maybe we should also think about creating proper fallback system, so if you do not have one translation, it's trying to get first fallback, second fallback, etc.

markitosgv:
Maybe we could inser a third fallbackLocale argument in proxyCurrentLocaleTranslation?

```php
<?php

/*
 * This file is part of the KnpDoctrineBehaviors package.
 *
 * (c) KnpLabs <http://knplabs.com/>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */

namespace Knp\DoctrineBehaviors\Model\Translatable;

use Doctrine\Common\Collections\ArrayCollection;

/**
 * Translatable trait.
 *
 * Should be used inside entity, that needs to be translated.
 */
trait TranslatableMethods
{
    /**
     * Returns collection of translations.
     *
[...]

PR #191, opened 2015-03-12
Add Callable function to override default language

jerome-fix:
refs #182

### DIFF
--- a/config/orm-services.yml
+++ b/config/orm-services.yml
@@ -56,7 +56,7 @@ services:
     knp.doctrine_behaviors.translatable_subscriber.default_locale_callable:
         class:           "%knp.doctrine_behaviors.translatable_subscriber.default_locale_callable.class%"
         arguments:
-            - "@service_container" # lazy request resolution
+            - %locale%
         public:  false
 
     knp.doctrine_behaviors.softdeletable_subscriber:
--- a/config/orm-services.yml
+++ b/config/orm-services.yml
@@ -3,6 +3,7 @@ parameters:
     knp.doctrine_behaviors.reflection.is_recursive: true
     knp.doctrine_behaviors.translatable_subscriber.class: Knp\DoctrineBehaviors\ORM\Translatable\TranslatableSubscriber
     knp.doctrine_behaviors.translatable_subscriber.current_locale_callable.class: Knp\DoctrineBehaviors\ORM\Translatable\CurrentLocaleCallable
+    knp.doctrine_behaviors.translatable_subscriber.default_locale_callable.class: Knp\DoctrineBehaviors\ORM\Translatable\DefaultLocaleCallable
     knp.doctrine_behaviors.translatable_subscriber.translatable_trait: Knp\DoctrineBehaviors\Model\Translatable\Translatable
     knp.doctrine_behaviors.translatable_subscriber.translation_trait: Knp\DoctrineBehaviors\Model\Translatable\Translation
     knp.doctrine_behaviors.translatable_subscriber.translatable_fetch_method: LAZY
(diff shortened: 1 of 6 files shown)


src/ORM/Translatable/DefaultLocaleCallable.php
```diff
@@ -0,0 +1,27 @@
...
+use Symfony\Component\DependencyInjection\Container;
+
+/**
+ * @author    Jérôme Fix <jerome.fix@zapoyok.info>
+ */
+class DefaultLocaleCallable
+{
+    private $container;
+
+    public function __construct(Container $container)
```
docteurklein: you should inject the parameter directly.

config/orm-services.yml
```diff
@@ -51,6 +53,12 @@ services:
             - "@service_container" # lazy request resolution
         public:  false
 
+    knp.doctrine_behaviors.translatable_subscriber.default_locale_callable:
+        class:           "%knp.doctrine_behaviors.translatable_subscriber.default_locale_callable.class%"
+        arguments:
+            - "@service_container" # lazy request resolution
```
docteurklein: - "%locale%"

why lazy? the parameter bag is frozen at build time, isn't it ?

jerome-fix: Something like that ? 

```yaml
    knp.doctrine_behaviors.translatable_subscriber.default_locale_callable:
        class:           "%knp.doctrine_behaviors.translatable_subscriber.default_locale_callable.class%"
        arguments:
            - %locale%
        public:  false
````

````php
class DefaultLocaleCallable
{
    private $locale;

    public function __construct($locale)
    {
        $this->locale = $locale;
    }

    public function __invoke()
    {
        return $this->locale;
    }
}
```

docteurklein:
exactly!

docteurklein:
(waiting for CI) and I'll merge

jerome-fix:
@docteurklein could you merge this pull request or explain me what i should modify or adapt ?

docteurklein:
thanks :)

jerome-fix:
You welcome

Le 2015-03-23 09:25, Florian Klein a écrit :

Rvanlaak:
This is nice, but can you add some documentation or a code sample about how to override the default locale? We get `nl_NL` as default locale, but we never make use of the full country locale anywhere. Did assume this was set to `en` as this is our framework fallback locale, but the `TranslatorSubscriber` thinks different about it.

MAXakaWIZARD:
@jerome-fix @Rvanlaak  As I see, listeners are called, but finally locale gets determined as `en`, while it is set to `ru` in my config (in Translatable currentLocale is finally `null` and defaultLocale is `en`)

MAXakaWIZARD:
Working solution is:
```php
/**
     * @ORM\PostUpdate
     * @ORM\PostPersist
     */
    public function updateSlug(LifecycleEventArgs $eventArgs)
    {
        $translatable = $this->getTranslatable();
        $em = $eventArgs->getEntityManager();
        $repo = $em->getRepository(get_class($translatable));
        //reload translatable to have current locale set
        $id = $translatable->getId();
        $em->detach($translatable);
        $translatable = $repo->find($id);

        $oldSlug = $translatable->getSlug();
        $translatable->generateSlug();

        if ($oldSlug !== $translatable->getSlug()) {
            $repo->updateSlug($translatable);
        }
    }
```
But I think it is quite cumbersome, maybe anyone can suggest better solution?

MAXakaWIZARD:
Conclusion is: locale setting does not work for newly created objects. But this can be fixed by adding subscription to `prePersist` lifecycle callback. @docteurklein what do you think? I can make a PR for it.

PR #193, opened 2015-03-18
Handle overridden classes for Translatable 

NV-iloweb:
This enable to override translatable/translation classes and the use of mapped superclasses in mappings.
Note : get_called_class was introduced in PHP 5.3


NV-iloweb:
Need to do some more check about mapped superclasses

PR #196, opened 2015-03-25
[Translatable] method setTranslations added

wolfwolker:
Added setter on translatableMethods trait.
Usefull to set multiple translations at same time. (Example: using nelmio/alice to load fixtures of translatable entities)

### DIFF
--- a/src/Model/Translatable/TranslatableMethods.php
+++ b/src/Model/Translatable/TranslatableMethods.php
@@ -56,6 +56,28 @@ trait TranslatableMethods
     }
 
     /**
+     * Setter for translations collection
+     * 
+     * @param ArrayCollection|Array $translations
+     *
+     * @return $this
+     */
+    public function setTranslations($translations)
+    {
+        if (!$translations instanceof \Traversable && !is_array($translations)) {
+            throw new \InvalidArgumentException('$translations parameter must be iterable');
+        }
+
+        $this->translations = new ArrayCollection();
+        
+        foreach ($translations as $translation) {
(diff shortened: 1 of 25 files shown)


src/Model/Translatable/TranslatableMethods.php
```diff
@@ -56,6 +56,28 @@ public function addTranslation($translation)
...
+     *
+     * @return $this
+     */
+    public function setTranslations($translations)
+    {
+        if (!$translations instanceof \Traversable && !is_array($translations)) {
+            throw new \InvalidArgumentException('$translations parameter must be iterable');
+        }
+
+        $this->translations = new ArrayCollection();
```
docteurklein: woujdn't this break persistent collections? Doctrine would get lost.

docteurklein:
Can you add a test case please?

src/Model/Translatable/TranslatableMethods.php
wolfwolker: I don't get you. Why?

docteurklein: I'm not sure doctrine will like to replace a persistent collection with another collection. That's why I'd like to see an integration test.

Einenlum: Indeed, I think you need to empty the current collection and replace its content.

TomasVotruba:
Replaced by https://github.com/KnpLabs/DoctrineBehaviors/pull/460, that is rebased on current `master` where so much has changed since.

Thank you for your contribution :+1:

PR #215, opened 2015-05-21
Function documented to receive null

crazycodr:
The documentation states that we should be able to set null to the deltedAt date which makes sense to allow "undeletion"

### DIFF
--- a/src/Model/SoftDeletable/SoftDeletableMethods.php
+++ b/src/Model/SoftDeletable/SoftDeletableMethods.php
@@ -85,7 +85,7 @@ trait SoftDeletableMethods
      *
      * @return $this
      */
-    public function setDeletedAt(\DateTime $date)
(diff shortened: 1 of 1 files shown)


crazycodr:
We need this at work for our project to continue, please consider merging ASAP.

crazycodr:
My changes are passing but the the whole package has other errors. Can we still approve this and merge, thanks?

akovalyov:
@crazycodr the phpdoc is invalid there, thank you for pointing this out. However, for "undeletion" functionaity you can use a method which is made [exactly for that](https://github.com/KnpLabs/DoctrineBehaviors/blob/977677d6645185b990607c136effefe08ce9b198/src/Model/SoftDeletable/SoftDeletableMethods.php#L32)

akovalyov:
Closed by #220.

PR #216, opened 2015-05-23
Attemt to put vendor folder to cache to prevent composer failures

akovalyov:


akovalyov:
Not possible to remove ```sudo``` usage atm, see http://docs.travis-ci.com/user/workers/container-based-infrastructure/

akovalyov:
Quick summary of the PR
1. Removed ```sudo``` usage from the travis config. It allows to use container-based travis infrastructure and they claim that in this case [builds are faster](http://docs.travis-ci.com/user/workers/container-based-infrastructure/) and as a bonus - it allows to use [caching](http://docs.travis-ci.com/user/caching/)
2. Added caching of Composer cache dir. As a result, composer deps are loaded from the fs cache which is
  1. faster (it allowed to reduce the build time from [1.5 hours](https://travis-ci.org/KnpLabs/DoctrineBehaviors/builds/63788944)-[2 hours](https://travis-ci.org/KnpLabs/DoctrineBehaviors/builds/63953160) as in last builds to [40 minutes](https://travis-ci.org/KnpLabs/DoctrineBehaviors/builds/63979785) and I think the build time will be lower when all deps will be cached.
  2. we won't hit a situation when Github API request limit is ended and the deps can not be installed. like in [this build job](https://travis-ci.org/KnpLabs/DoctrineBehaviors/jobs/63942536)

cursedcoder:
cool

PR #224, opened 2015-06-10
[Sluggable] Separate slug transliterator

sergeyfedotov:
This feature allows to use custom slug transliteration class.

### DIFF
--- a/src/Model/Sluggable/SluggableMethods.php
+++ b/src/Model/Sluggable/SluggableMethods.php
@@ -68,7 +68,7 @@ trait SluggableMethods
      * @param $values
      * @return mixed|string
      */
-    private function generateSlugValue($values)
+    private function generateSlugValue($values, TransliteratorInterface $transliterator)
     {
         $usableValues = [];
         foreach ($values as $fieldName => $fieldValue) {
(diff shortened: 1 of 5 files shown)


src/Model/Sluggable/SluggableMethods.php
```diff
@@ -68,7 +68,7 @@ public function getSlug()
      * @param $values
      * @return mixed|string
      */
-    private function generateSlugValue($values)
+    private function generateSlugValue($values, TransliteratorInterface $transliterator)
```
docteurklein: I'm not sure i like the idea of injecting the transliterator during the method call.

akovalyov:
Any activity/holywars on it?

MAXakaWIZARD:
@akovalyov You could just override `generateSlug` method and use any transliterator you want, I did it in such way before my PR #162 was accepted and merged.

TomasVotruba:
Is this feature still needed? Could you rebase if so?

TomasVotruba:
As the service is static hard-coded in object and serves only as default implementaion,
overriding it is the way to go:
https://github.com/KnpLabs/DoctrineBehaviors/pull/224#issuecomment-125553897

Closing as resolved

PR #228, opened 2015-06-17
Change mysite.com with example.com

bocharsky-bw:
I think in examples better to use `example.com` domain


docteurklein:
indeed, thanks!

Issue #208, opened 2015-05-06
[Timestampable/Blameable] [RFC] Overhaul of "Trackable" behaviors (Timestampable/Blameable) + inclusion of IpTraceable


lemoinem:
Hello,

This issue/RFC will include several discussion, but I think it's easier to start with the big picture and break down the discussion if necessary.

I am currently moving away from Stof/Gedmo's Doctrine-Extensions since the bundle doesn't seem to be maintained anymore, and I noticed a few things:
- Timestampable and Blameable are not consistent with each other (Timestampable calls a method in the entity while Blameable relies on a completely external listener)
- They aren't as flexible as in Gedmo (for example the `on="change"` behavior)
- IpTraceable does not exists in Behaviors (yet?), It's in Gedmo, although the support hasn't been included in Stof's Bundle yet (stof/StofDoctrineExtensionsBundle#233)

I would like to propose a complete overhaul (separated in a few PRs) solving these issues. However, before starting to work on it, I would like to know if you are interested in such PRs and, if yes, there are a few design choices I would like to discuss.
## Consistency

I propose All "Trackable" behaviors use a single Listener.
The Listener would be provided a set of `Callable`s, each providing an associative array of meta-data.
On `prePersist` or `preUpdate` events, the sets of all meta-data would be merged together and injected in a new (custom) `LifecycleEvent` and given to a method of the behavior Trait. This method would be responsible for actually updating the Entity accordingly.

I believe this scenario to fit the philosophy of other behaviors available in the bundle and to be easily extensible if needed.
## Flexibility

Since the whole Behaviors system is based on traits and Annotation would be a major disruption, we need to achieve the flexibility in another way.

Granted the consistency mechanism is accepted, I propose the Trackable Trait provides a method `updateConditions` which would return a list of tupples: `[tag, event, field, value]`. The `event`, `field` and `value` are based on [Gedmo's system](https://github.com/Atlantic18/DoctrineExtensions/blob/master/doc/timestampable.md#timestampable-annotations). So `event` could be either of `create`, `update` or `change` (Maybe `update` and `change` could be merged?), `field` would be a (set of) field(s) to watch and `value` a value to wait the given field to change to.

The list of conditions would be evaluated by the Listener and appropriate tags would be added to the event meta-data.
The default implementation for each behavior trait would simply provide the default `[["create", "create"], ["update", "update"]]` (These two cases could be abbreviated to `[["create"], ["update"]]` or even `["create", "update"]`.
[...]

PR #233, opened 2015-07-13
[Blameable/Timestampable] Consistency of Trackable behaviors

lemoinem:
Hello,

This PR reprensents the consistency part of RFC #208.

This PR intends to propose a new architecture to simplify _Trackable_ behaviors. _Trackable_ behaviors include _Blameable_ and _Timestampable_. The main idea is that the _Entity_ becomes responsible for tracking meta-data regarding its own creation, changes, and deletion (the last part is intended to be used along _SoftDeletable_ as were _Blameable_ and _Timestampable_).

The new architecture uses only a single Subscriber on `preCreate`, `preUpdate` and `preRemove` for all _Trackable_ behaviors. This new Subscriber is more generic and improve consistency among the implementations.

The _Trackable_ Subscriber is given a collection of `Tracker`s, and is responsible for gathering the metadata for a given Entity each time an event is triggered. A `Tracker` can be developed by implementing the provided `TrackerInterface` (recommended) or by providing a triplet (`name`, `predicate`, `generator`).

If the full symfony framework is used (Bundle, Extensions and Compiler passes), the new `knp.doctrine_behaviors.tracker` tag can be used to easily add custom trackers.

Each `Tracker` is responsible for fetching/generating (generally speaking, providing) some metadata. Each `Tracker` is responsible for deciding if it will provide its own metadata for a given Event.

The _Entity_ itself is responsible for processing the metadata made available to it and storing or processing it adequately.

I consider the architecture to respect the SRP because the Entity's responsibility is akin to LifeCycleCallbacks which already exists in Doctrine. This way, the new responsibility is not a new burden, simply an extension of the Event System already provided by Doctrine.

I had to keep the _Blameable_ and _Timestampable_ Subscribers on `loadClassMetadata` to ensure the mapping of the fields provided in the Traits.

I introduce two BCB:
1. `updateTimestamps` method in the _Timestampable_ Trait is removed. However, if you think this is a major modification, I'm sure we could work something out to ensure the methods is preserve. Although I think this would hurt the new-found consistency.
2. Entities using either the _Blameable_ or _Timestampable_ Trait must now also use the _Trackable_ trait. I failed to find a way to avoid this, however, I'm completely opened to feedback or idea on this point.

Tests and documentation are up-to-date.

I added a dependency on `symfony/property-access` for the tests only, although I can easily work around it if need be.

Thanks for your time reading and providing feedback on this PR.

TODO:
- [x] ~~Write tests for Trackable~~
- [x] ~~Update tests for Blameable Subscriber~~
- [x] ~~Write tests for Blameable Trait~~
- [x] ~~Update tests for Timestampable Subscriber~~
- [x] ~~Write tests for Timestampable Trait~~
- [x] ~~Write doc~~
- [x] ~~Write a new `CompilerPass` to allow defining new `Tracker`s by defining a tag on a service implementing `TrackerInterface`~~


lemoinem:
@docteurklein : Sorry to bother you directly, but the PR has been opened for three weeks now, without feedback.

I know it's still a WIP, but could I ask you to take a look at it and tell me if I'm heading in the right direction?

docteurklein:
I'm so sorry, I didn't have time to answer you. Saw the PR already, and the idea seems OK to me.
Not sure if the added value will worth the complication.

lemoinem:
OK, I've (re)written all involved test.
NB: I added a dev dependency on `symfony/property-access`

The code should be in working order now.
Granted I plugged the service (Subscribers, Trackers & Cie) correctly.

I will review these while writting the documentation and the CompilerPass for the Tracker tag.

Any feedback welcome, although I gather the documentation might help understand the new structure a bit better.

lemoinem:
Here comes the documentation.

lemoinem:
And the compiler pass...

@docteurklein My next goal would be to rebase and reorganize my commits but the code is final.

I will probably rewrite the head message to reflect the last upgrades...

Could you review the final architecture and documentation.

As far as I see, there is an increase in consistency, simplicity and flexibility.

I checked whether the Blameable and Timestampable traits could avoid requiring the manual use of the Trackable trait, however, if I want to let users easily change the timestampable trait, it seemed better this way.

On the other hand, the Trackable behavior could be conditioned to an interface, since the trait only declares methods, but it's currently impossible to provide an interface through a trait. So classes using the Blameable or Timestampable trait would still need to explicitly reference the trackable trait.

lemoinem:
Travis fail comes from the fact that Doctrine's `dev-master` branch now requires PHP 5.5+.
Therefore the  5.4 builds using Doctrine's `dev-master` fail.

lemoinem:
PS: Head message rewritten. @docteurklein, if you have some time, the PR is ready for review, including CS and commit squashing/rebasing.

docteurklein:
@lemoinem thanks for your work! will review ASAP.

lemoinem:
I just thought about it, it might be possible to allow Blameable and Timestampable traits not to require the Trackable trait by increasing the coupling between the three implementations even further. (I'm sure if it is a good idea...)

However, I believe the coupling between the traits can't really be increased, it may create issue with entities using both traits. Creating a new trait encompassing both implementations at once might be created to this end. (i.e. `BlameableAndTimestampable`).

How about using interfaces instead of Traits to decide whether an entity should be processed by a subscribers? It seems to be a better OOP Practice too.

docteurklein:
I agree using interfaces is a better idea. It would resolve and simplify so much prolbems in this lib. I think there is a PR actually doing this.

I had no time to manage it yet.

lemoinem:
Indeed, the PR is #241 (and there is some more discussion in #237). If you think it would help, I could try to write PRs for the missing behavior interfaces. Then I would rewrite this PR using interface instead of traits (would be much better anyway, IMO).

TomasVotruba:
Closing as https://github.com/KnpLabs/DoctrineBehaviors/pull/233#issuecomment-139530641

In respect with Occams' razor, use smaller separated concepts `Blameable` and `Timestampable` instead

Issue #234, opened 2015-07-15
[Sluggable] How could I generate unique slug with sluggable Trait?


bocharsky-bw:
I try to use Symfony Validator Callbacks to check whether slug is unique manually and add violation if not, but this unsuccessful finished because validators triggered before slug was generated. Any suggestions?

akovalyov:
@bocharsky-bw could you provide a little bit more info(code I mean) about what are you trying to achieve?

bocharsky-bw:
@akovalyov Thank you for your response! In [docs](https://github.com/KnpLabs/DoctrineBehaviors#sluggable) I see **uniqueness is not guaranteed** and it's true. So let me simplify my question a bit:
How could I generate an unique slug with sluggable Trait?

akovalyov:
what about something ike
```php
    public function getSluggableFields()
    {
        return ['id',  'name' ];
    }
```
in your entity?

bocharsky-bw:
@akovalyov Yeah, it's a nice idea, in some projects it will be work, but what if I need slug without id in it. Could I get some behavior like:

```
some-entity-slug
some-entity-slug-1
some-entity-slug-2
// ... etc.
```

I need to add a counter at the end of the slug if such slug already used. How can I do it?

akovalyov:
@bocharsky-bw I am not sure if you can do it with Sluggable and I am not sure that you *should* do it with Sluggable. You will need to interact with your DB and it should not be done on the Model level. 
I have another question. Why do you need a unique slug? You already have an id, which is unique.

bocharsky-bw:
Well, in a business world we get some tasks that we must do as the customer wants. Also I think you know that many sites has an unique slug without ID in it )

I agree with you to do it with Sluggable is wrong idea. But maybe we should to create an event listener for it where we could interact with DB?

Anyway, a tool which will be solved such business task seems to be an useful, in my opinion.

akovalyov:
Wait. If we are talking about business world and business tasks - I have never heard about the business task to be able to generate unique slug in DB. 
 If we are talking about business task for human-friendly URLs - that is not the task of Sluggable, because we do not care about the strategy for unique URLs. We only generate slugs.

bocharsky-bw:
OK, what about business task for human-friendly URLs... OK, that is not the task of Sluggable directly. But we could add a tool to this repository which will solve this tasks (care about the strategy of unique URLs)! Why not? :)

[...]

PR #236, opened 2015-07-18
[Sluggable] Add feature to generate unique slug

bocharsky-bw:
Resolve issue #234 .

If the `SluggableMethods:: getRegenerateUniqueSlug()` method returns `true` - the `SluggableSubscriber` will generate unique slug for entity with counter at the end if slug already taken, for example:

### DIFF
--- a/config/orm-services.yml
+++ b/config/orm-services.yml
@@ -154,4 +154,3 @@ services:
             - "%knp.doctrine_behaviors.sluggable_subscriber.sluggable_trait%"
         tags:
             - { name: doctrine.event_subscriber }
-
--- a/src/Model/Sluggable/SluggableMethods.php
+++ b/src/Model/Sluggable/SluggableMethods.php
@@ -42,6 +42,16 @@ trait SluggableMethods
     }
 
     /**
+     * Returns whether or not the unique slug gets regenerated.
+     *
+     * @return bool
+     */
+    public function getRegenerateUniqueSlug()
+    {
+        return false;
+    }
+
+    /**
      * Sets the entity's slug.
      *
      * @param $slug
--- a/src/Model/Sluggable/SluggableMethods.php
+++ b/src/Model/Sluggable/SluggableMethods.php
@@ -46,7 +46,7 @@ trait SluggableMethods
      *
      * @return bool
      */
-    public function shouldGenerateUniqueSlugs()
+    public static function shouldGenerateUniqueSlugs()
     {
         return false;
     }
--- a/src/Model/Sluggable/SluggableMethods.php
+++ b/src/Model/Sluggable/SluggableMethods.php
@@ -46,7 +46,7 @@ trait SluggableMethods
      *
      * @return bool
      */
-    public function getRegenerateUniqueSlug()
+    public function shouldGenerateUniqueSlugs()
     {
         return false;
     }
--- a/src/ORM/Sluggable/SluggableSubscriber.php
+++ b/src/ORM/Sluggable/SluggableSubscriber.php
@@ -109,12 +109,11 @@ class SluggableSubscriber extends AbstractSubscriber
      */
     private function generateUniqueSlugFor($entity, EntityManager $em)
     {
-        $classMetadata = $em->getClassMetadata(get_class($entity));
         $slug = $entity->getSlug();
         $uniqueSlug = $slug;
 
         $i = 0;
-        while ($em->getRepository($classMetadata->getName())->matching(
+        while ($em->getRepository(get_class($entity))->matching(
             Criteria::create()
                 ->andWhere(Criteria::expr()->neq('id', $entity->getId()))
                 ->andWhere(Criteria::expr()->eq('slug', $uniqueSlug))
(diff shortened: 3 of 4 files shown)


bocharsky-bw:
I could add some docs if it's OK

src/Model/Sluggable/SluggableMethods.php
```diff
@@ -42,6 +42,16 @@ private function getRegenerateSlugOnUpdate()
     }
 
     /**
+     * Returns whether or not the unique slug gets regenerated.
+     *
+     * @return bool
+     */
+    public function getRegenerateUniqueSlug()
```
docteurklein: better name:  `shouldGenerateUniqueSlugs`

src/ORM/Sluggable/SluggableSubscriber.php
```diff
@@ -93,4 +102,26 @@ private function isSluggable(ClassMetadata $classMetadata)
...
         );
     }
+
+    /**
+     * @param $entity
+     * @param EntityManager $em
+     */
+    private function generateUniqueSlugFor($entity, EntityManager $em)
+    {
+        $classMetadata = $em->getClassMetadata(get_class($entity));
```
docteurklein: why using classMetadata?

You can directly do : `$em->getRepository(get_class($entity))`

src/Model/Sluggable/SluggableMethods.php
bocharsky-bw: Yes, you're right and thanks for your suggestion!
I want to be consistent with `getRegenerateSlugOnUpdate` only.
Anyway, the `shouldGenerateUniqueSlugs` name will be better.

src/ORM/Sluggable/SluggableSubscriber.php
bocharsky-bw: I first thought I should use it in this method. I removed it, thanks

src/ORM/Sluggable/SluggableSubscriber.php
```diff
@@ -93,4 +102,25 @@ private function isSluggable(ClassMetadata $classMetadata)
...
+     * @param $entity
+     * @param EntityManager $em
+     */
+    private function generateUniqueSlugFor($entity, EntityManager $em)
+    {
+        $slug = $entity->getSlug();
+        $uniqueSlug = $slug;
+
+        $i = 0;
+        while ($em->getRepository(get_class($entity))->matching(
```
bocharsky-bw: @docteurklein What do you think about `matching()` method in this case? Is it an overhead? Maybe better to use a QueryBuilder here?

docteurklein: good question, i never use criteria.

cursedcoder: @docteurklein is there any possibility when `entity` will be the proxy class?

bocharsky-bw: @docteurklein OK, I'll try to replace Criteria with QueryBuilder

bocharsky-bw:
@docteurklein What do you think about to move query building to the entity repository?

Developers could extends [SluggableRepository](https://github.com/bocharsky-bw/DoctrineBehaviors/blob/unique-slug/src/Repository/SluggableRepository.php) in they custom entity repository to override `isSlugUniqueFor()` method for custom implementation of unique slug checking. 
Also if developer do not extends `SluggableRepository` - this repo will be created automatically in subscriber by default.

config/orm-services.yml
```diff
@@ -153,4 +153,3 @@ services:
             - "%knp.doctrine_behaviors.sluggable_subscriber.sluggable_trait%"
         tags:
             - { name: doctrine.event_subscriber }
-
```
kuczek: Why have you removed that unit?

bocharsky-bw: Check [original file](https://github.com/bocharsky-bw/DoctrineBehaviors/edit/master/config/orm-services.yml), please. There're 2 empty new lines. I removed one.

src/ORM/Sluggable/SluggableSubscriber.php
```diff
@@ -93,4 +103,27 @@ private function isSluggable(ClassMetadata $classMetadata)
...
+     * @param $entity
+     * @param EntityManager $em
+     * @param ClassMetadata $classMetadata
+     */
+    private function generateUniqueSlugFor($entity, EntityManager $em, ClassMetadata $classMetadata)
+    {
+        $repo = $em->getRepository($classMetadata->name);
+        if (!$repo instanceof SluggableRepository) {
+            // Create default repository if entity one do not extends it
+            $repo = new SluggableRepository($em, $classMetadata);
```
docteurklein: I don't think it's a good idea to instanciate the repo by hand

bocharsky-bw: I don't like it too. What do you think is better to use in this case?

docteurklein: I wouldn't add this feature in the first place. I think unique urls are resolvable by adding `/{id}/{slug}`.

TomasVotruba:
Merged in #449, thank you for your work :+1:

PR #237, opened 2015-07-19
[Sluggable] Use interface instead of trait to determine sluggable entity

bocharsky-bw:


bocharsky-bw:
With interface it looks more clearer and readable than with trait. The sluggable entity should realize `SluggableInterface` in order to generate slug by `SluggableSubscriber`. Also with interface we get auto-completion by IDEs in `SluggableSubscriber` and other parts of code.
The main purpose of this PR is ability to easy override `Sluggable` trait with custom one if other developers need to realize their custom logic in order to create the slug. Also in this way we could define `$slug` property directly in entity by ourselves with custom annotation mapping.

docteurklein:
I tend to agree that interfaces are clearly better than our hacky trait duck typing.
I think we did this at fist to facilitate the end user life, but I don't think adding an interface would be so complicated.

So +1, it would remove all the complicated trait resolution and simplify the lib :)
It is however a BC break we should document.

bocharsky-bw:
I also like an idea to introduce an interface way for all traits. I could help and do it in this PR or create another one. What do you think?

What is the better way to avoid BC break? We could release a new major version `v2.0.0` for interface way to correctly avoid BC break I think.

docteurklein:
yes, a major release is the best way. it's not a complicated update anyway.

docteurklein:
yes a pr is welcome. look if you can get rid of traitanalyzer and this 'isRecursive' thing :-)

bocharsky-bw:
@docteurklein I removed unnecessary part of code that was replaced by interface implementation way. What do you think?

src/ORM/Sluggable/SluggableSubscriber.php
```diff
@@ -40,12 +27,12 @@ public function loadClassMetadata(LoadClassMetadataEventArgs $eventArgs)
             return;
         }
 
-        if ($this->isSluggable($classMetadata)) {
+        if (is_subclass_of($classMetadata->getName(), 'Knp\DoctrineBehaviors\Model\Sluggable\SluggableInterface')) {
```
lemoinem: Wouldn't you prefer basing yourself on the *Properties trait to trigger the mapping ? Using the interface might cause problems in case of inheritance (CF #317).

TomasVotruba:
Very nice job with PR split and minimal scope! It's easy for me to pick it up and integrate :) thank you

PR #242, opened 2015-07-30
[Timestampable] Use timestampable interface instead of trait

bocharsky-bw:


src/ORM/Timestampable/TimestampableSubscriber.php
```diff
@@ -43,42 +33,27 @@ public function loadClassMetadata(LoadClassMetadataEventArgs $eventArgs)
             return;
         }
 
-        if ($this->isTimestampable($classMetadata)) {
-            if ($this->getClassAnalyzer()->hasMethod($classMetadata->reflClass, 'updateTimestamps')) {
-                $classMetadata->addLifecycleCallback('updateTimestamps', Events::prePersist);
-                $classMetadata->addLifecycleCallback('updateTimestamps', Events::preUpdate);
-            }
+        if (is_subclass_of($classMetadata->getName(), 'Knp\DoctrineBehaviors\Model\Timestampable\TimestampableInterface')) {
```
lemoinem: Wouldn't you prefer basing yourself on the *Properties trait to trigger the mapping ? Using the interface might cause problems in case of inheritance (CF #317).

PR #246, opened 2015-08-10
[Tree] Replace recursive trait check with interface for tree model

bocharsky-bw:


docteurklein:
it broke the test suite.

src/ORM/Tree/TreeSubscriber.php
```diff
@@ -43,13 +31,26 @@ public function loadClassMetadata(LoadClassMetadataEventArgs $eventArgs)
             return;
         }
 
-        if ($this->isTreeNode($classMetadata)) {
-
+        if (is_subclass_of($classMetadata->getName(), 'Knp\DoctrineBehaviors\Model\Tree\NodeInterface')) {
```
docteurklein: better use `instanceof`

bocharsky-bw: Yes, but `instanceof` works with objects, but I have a string of class name

docteurklein: ho, you're right sorry.

lemoinem: Wouldn't you prefer basing yourself on the *Properties trait to trigger the mapping ? Using the interface might cause problems in case of inheritance (CF #317).

PR #248, opened 2015-08-11
[Softdeletable] Make soft deletable subscriber to work with interface instead of traits

bocharsky-bw:
Make subscriber to work with interface instead of traverse traits recursively for soft deletable model

### DIFF
--- a/src/ORM/SoftDeletable/SoftDeletableSubscriber.php
+++ b/src/ORM/SoftDeletable/SoftDeletableSubscriber.php
@@ -11,15 +11,11 @@
 
 namespace Knp\DoctrineBehaviors\ORM\SoftDeletable;
 
-use Knp\DoctrineBehaviors\Reflection\ClassAnalyzer;
-
-use Knp\DoctrineBehaviors\ORM\AbstractSubscriber;
-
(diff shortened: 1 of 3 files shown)


src/ORM/SoftDeletable/SoftDeletableSubscriber.php
```diff
@@ -48,9 +35,10 @@ public function onFlush(OnFlushEventArgs $args)
         $em  = $args->getEntityManager();
         $uow = $em->getUnitOfWork();
 
+        /** @var SoftDeletableInterface $entity */
         foreach ($uow->getScheduledEntityDeletions() as $entity) {
             $classMetadata = $em->getClassMetadata(get_class($entity));
-            if ($this->isSoftDeletable($classMetadata)) {
+            if (is_subclass_of($classMetadata->getName(), 'Knp\DoctrineBehaviors\Model\SoftDeletable\SoftDeletableInterface')) {
```
lemoinem: Wouldn't you prefer basing yourself on the *Properties trait to trigger the mapping ? Using the interface might cause problems in case of inheritance (CF #317).

PR #249, opened 2015-08-11
[Sortable] Make subscriber to work with interface instead of traits for sortable model

bocharsky-bw:


src/Model/Sortable/Sortable.php
```diff
@@ -2,49 +2,12 @@
 
 namespace Knp\DoctrineBehaviors\Model\Sortable;
 
+/**
+ * Trait Sortable
```
stloyd: Quite useless IMO.

src/Model/Sortable/SortableInterface.php
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Interface SortableInterface
+ */
+interface SortableInterface
+{
+    /**
+     * Get sort.
+     *
+     * @return mixed
```
stloyd: This is `int`.

src/Model/Sortable/SortableInterface.php
```diff
@@ -0,0 +1,39 @@
...
+     */
+    public function getSort();
+
+    /**
+     * Set sort.
+     *
+     * @param mixed $sort
```
stloyd: Same.

src/Model/Sortable/SortableInterface.php
```diff
@@ -0,0 +1,39 @@
...
+     *
+     * @return $this
```
stloyd: Fluent interfaces are evil: http://ocramius.github.io/blog/fluent-interfaces-are-evil/

stloyd:
TBH. what's the point of spliting properties & methods? This leads IMO to more problems than it solves...

bocharsky-bw:
@stloyd Well, the problem is that PHP allow override trait's methods and don't allow to override trait's properties. So when I want to redefine trait's properties, for example, to change or add some extra annotation I have a fatal error.

src/Model/Sortable/SortableInterface.php
bocharsky-bw: Thanks, I fixed it

src/Model/Sortable/Sortable.php
bocharsky-bw: Well, we could to add more description here. Btw, it also needs some descriptions for docs. There is missing behaviour description.

NiR-:
:+1: :+1: :+1: 
@bocharsky-bw Thanks for this PR. Like said by stloyd, can you remove comments on top of traits and interfaces, or add more details in it (to make them real phpdoc) ?

bocharsky-bw:
@NiR- Done! Added some description in https://github.com/bocharsky-bw/DoctrineBehaviors/commit/6987d04c6b1d9eca49ae1e90c2d8330abe537c6f to all sortable traits

bocharsky-bw:
Seems some checks has error and I have no idea why... but it's not my fault, isn't it?

NiR-:
@bocharsky-bw Nope, it does not seems to be your fault :) But it will need another PR to fix those errors: some builds seem to fail due to a higher php constraints on `doctrine/common:dev-master`, and others due to too many requests to github API.

bocharsky-bw:
OK, when it'll be fixed I could rebase this PR to ensure all checks passed.

src/ORM/Sortable/SortableSubscriber.php
```diff
@@ -43,35 +33,21 @@ public function loadClassMetadata(LoadClassMetadataEventArgs $eventArgs)
             return;
         }
 
-        if ($this->isSortable($classMetadata)) {
-
+        if (is_subclass_of($classMetadata->getName(), 'Knp\DoctrineBehaviors\Model\Sortable\SortableInterface')) {
```
lemoinem: Wouldn't you prefer basing yourself on the *Properties trait to trigger the mapping ? Using the interface might cause problems in case of inheritance (CF #317).

PR #262, opened 2015-10-15
Timestampable - add db field type parameter

lopsided:
We came across this issue when working with a Postgres database. The `datetime` field in pg ignores the timezone which was causing us some headaches. This PR allows you to define the field type as a parameter rather than (as we have done in our current project) overriding the whole subscriber class - which, incidentally, has a number of private methods and properties which make for an unnecessarily large override file. Have added some docs too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -448,6 +448,19 @@ so that when you try to call `getName` (for example) it will return you the tran
 
 ```
 
+If you wish to change the doctrine type of the database fields that will be created for timestampable models you can
+set the following parameter like so:
+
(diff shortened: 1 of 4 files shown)


NiR-:
:+1: @lopsided Tests are broken, can you fix it ?

lopsided:
Whoops sorry missed that. Have fixed the test and squashed. Let me know if I've missed anything else
@NiR-

docteurklein:
thanks!

PR #265, opened 2015-11-10
Language only fallback

DerekRoth:
When working with **language_COUNTRY** locales, it is common to fallback to language only translations. However, Translatable could only fallback to the default locale. This PR fixes this.

Maybe it has to be enhanced, as it uses the same **$fallbackToDefault** parameter as the one that already exists, but its usage isn't quite the same. If you previously relied on the fallback of **fr_FR** to **en** instead of **fr**, you will have a BC break (currently no unit tests show this).

See the code of the [Symfony translation component](https://github.com/symfony/translation/blob/master/Translator.php#L392) for reference.

### DIFF
--- a/src/Model/Translatable/TranslatableMethods.php
+++ b/src/Model/Translatable/TranslatableMethods.php
@@ -104,10 +104,9 @@ trait TranslatableMethods
         }
 
         if ($fallbackToDefault) {
-            if ($fallbackLocale = $this->computeFallbackLocale($locale)) {
-                if ($translation = $this->findTranslationByLocale($fallbackLocale)) {
-                    return $translation;
-                }
+            if (($fallbackLocale = $this->computeFallbackLocale($locale))
+                && ($translation = $this->findTranslationByLocale($fallbackLocale))) {
+                return $translation;
             }
 
             if ($defaultTranslation = $this->findTranslationByLocale($this->getDefaultLocale(), false)) {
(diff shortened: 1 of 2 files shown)


EmmanuelVella:
:+1:

DerekRoth:
Sorry i messed up with my branches a minute ago. Anyway, this could be merged if you can review the code or discuss it.

src/Model/Translatable/TranslatableMethods.php
```diff
@@ -103,8 +103,16 @@ protected function doTranslate($locale = null, $fallbackToDefault = true)
             return $translation;
         }
 
-        if ($fallbackToDefault && $defaultTranslation = $this->findTranslationByLocale($this->getDefaultLocale(), false)) {
-            return $defaultTranslation;
+        if ($fallbackToDefault) {
+            if ($fallbackLocale = $this->computeFallbackLocale($locale)) {
+                if ($translation = $this->findTranslationByLocale($fallbackLocale)) {
```
NiR-: You can group those two if.

NiR-:
:+1:

src/Model/Translatable/TranslatableMethods.php
DerekRoth: I actually tried that but it then complains about the $fallbackLocale not being defined. And it would make the code less readable too in my opinion.

DerekRoth: It works better after adding some parenthesis.

docteurklein:
thanks!

PR #266, opened 2015-11-17
Make Behaviors configurable.

NiR-:
You can now enable/disable behaviors you don't want using the bundle config.
By default, all behaviors are enabled. As soon as `knp_doctrine_behaviors` is
specified, all behaviors are disabled and the developer need to specify which
ones he want to enable.

This fix conflict issues, for example when a `setCurrentLocale` method is
defined by another third-party library/bundle.

This way does not introduce BC breaks, but only a new feature. The best would
be to add some interfaces and use them in subscribers. But that would imply
a major release... Maybe for the next one :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,19 @@ Make sure to activate them by reading the [Subscribers](#subscribers) section.
 ##Installation
 ```composer require knplabs/doctrine-behaviors:~1.1```
 
+## Configuration
+By default, when integrated with Symfony, all subscribers are enabled (if you don't specify any configuration for the bundle).
+But you can enable behaviors you need in a whitelist manner:
+```yaml
+knp_doctrine_behaviors:
+    blameable:      false
+    geocodable:     ~     # Here null is converted to false
+    loggable:       ~
+    sluggable:      true
+    soft_deletable: true
+    # All others behaviors are disabled
+```
+
 <a name="subscribers" id="subscribers"></a>
 ## Subscribers
 
--- a/src/Bundle/DependencyInjection/DoctrineBehaviorsExtension.php
+++ b/src/Bundle/DependencyInjection/DoctrineBehaviorsExtension.php
@@ -3,6 +3,7 @@
 namespace Knp\DoctrineBehaviors\Bundle\DependencyInjection;
 
 use Symfony\Component\Config\FileLocator;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Extension\Extension;
 use Symfony\Component\DependencyInjection\Loader\YamlFileLoader;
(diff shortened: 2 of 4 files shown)


src/Bundle/DependencyInjection/CompilerPass/DisableBehaviorsPass.php
```diff
@@ -0,0 +1,36 @@
...
+            'timestampable',
+            'translatable',
+            'tree',
+        ];
+
+        foreach ($behaviors AS $behavior)  {
+            if (true === $builder->getParameter(sprintf('knp_doctrine_behaviors.%s.enabled', $behavior))) {
+                continue;
+            }
+            var_dump($behavior);
```
docteurklein: missing var_dump

README.md
```diff
@@ -35,7 +48,7 @@ If you use symfony2, you can easily register them in:
 
 - *Recommended way:*
 
-Add to AppKernel
+Add to AppKernel, and make sure it is declared **before** `DoctrineBundle` (if you want disable some).
```
docteurklein: would it be possible to make it order independant (by playing with compiler pass priority f.e)?

docteurklein:
I think now we have a bundle, we should load the orm-services.yml file automatically, and we would thus have complete control over definitions.

The way you implemented it right now is to let the user include orm-services.yml,  and then remove the unwanted definitions (thus the constraint with bundle registration order).
If you go the other way around ( generate definitions on the fly only for wanted behaviors), all the problems are gone.

Thoughts?

docteurklein:
EDIT: sorry, the yml file is already included by us.
Anyway, I think we simply could remove the definition in the extension class directly, instead of using a compiler pass.

src/Bundle/DependencyInjection/DoctrineBehaviorsExtension.php
```diff
@@ -16,5 +17,21 @@ public function load(array $configs, ContainerBuilder $container)
     {
         $loader = new YamlFileLoader($container, new FileLocator(__DIR__.'/../../../config'));
         $loader->load('orm-services.yml');
+
+        $configuration = new Configuration();
+        $config = $this->processConfiguration($configuration, $configs);
+
+        foreach ($config AS $behavior => $enabled) {
```
docteurklein: lowercase `AS`

src/Bundle/DependencyInjection/DoctrineBehaviorsExtension.php
```diff
@@ -16,5 +17,21 @@ public function load(array $configs, ContainerBuilder $container)
     {
         $loader = new YamlFileLoader($container, new FileLocator(__DIR__.'/../../../config'));
         $loader->load('orm-services.yml');
+
+        $configuration = new Configuration();
+        $config = $this->processConfiguration($configuration, $configs);
+
+        foreach ($config as $behavior => $enabled) {
+            $container->setParameter(sprintf('knp_doctrine_behaviors.%s.enabled', $behavior), $enabled);
```
docteurklein: I don't think we need to create a parameter to store if enabled. The presence of the service is enough.

src/Bundle/DependencyInjection/DoctrineBehaviorsExtension.php
```diff
@@ -16,5 +17,21 @@ public function load(array $configs, ContainerBuilder $container)
...
+            $container->removeDefinition(sprintf('knp.doctrine_behaviors.%s_subscriber', $behavior));
```
docteurklein: wouldn't that remove **all** the definitions, even if we put `true` ?

docteurklein:
LGTM

akovalyov:
:+1:

PR #284, opened 2016-01-25
Microseconds

boekkooi:
Hi there,

So I was writing some tests and found out that the Timestampable, SoftDeletable and Loggable behaviours where ignoring microseconds which was causing some issue in my enviroment.

This PR adds microseconds to the Timestampable, SoftDeletable and Loggable behaviours. 

Let me know what you think.

Cheers,
Warnar

P.S. the `$dateTime->setTimezone` is needed because for some reason the unix timestamp changes based on the timezone in PHP.

### DIFF
--- a/src/Model/Timestampable/TimestampableMethods.php
+++ b/src/Model/Timestampable/TimestampableMethods.php
@@ -65,10 +65,14 @@ trait TimestampableMethods
      */
     public function updateTimestamps()
     {
+        // Create a datetime with microseconds
+        $dateTime = \DateTime::createFromFormat('U.u', sprintf('%.6F', microtime(true)));
+        $dateTime->setTimezone(new \DateTimeZone(date_default_timezone_get()));
+
         if (null === $this->createdAt) {
-            $this->createdAt = new \DateTime('now');
+            $this->createdAt = $dateTime;
         }
 
-        $this->updatedAt = new \DateTime('now');
+        $this->updatedAt = $dateTime;
     }
 }
--- a/src/Model/Loggable/Loggable.php
+++ b/src/Model/Loggable/Loggable.php
@@ -27,7 +27,7 @@ trait Loggable
         foreach ($changeSets as $property => $changeSet) {
             for ($i = 0, $s = sizeof($changeSet); $i < $s; $i++) {
                 if ($changeSet[$i] instanceof \DateTime) {
-                    $changeSet[$i] = $changeSet[$i]->format("Y-m-d H:i:s");
+                    $changeSet[$i] = $changeSet[$i]->format("Y-m-d H:i:s.u");
                 }
             }
 
--- a/src/Model/SoftDeletable/SoftDeletableMethods.php
+++ b/src/Model/SoftDeletable/SoftDeletableMethods.php
@@ -23,7 +23,7 @@ trait SoftDeletableMethods
      */
     public function delete()
     {
-        $this->deletedAt = new \DateTime();
+        $this->deletedAt = $this->currentDateTime();
     }
 
     /**
@@ -42,7 +42,7 @@ trait SoftDeletableMethods
     public function isDeleted()
     {
         if (null !== $this->deletedAt) {
-            return $this->deletedAt <= (new \DateTime());
+            return $this->deletedAt <= $this->currentDateTime();
         }
 
         return false;
@@ -90,5 +90,18 @@ trait SoftDeletableMethods
         $this->deletedAt = $date;
 
         return $this;
+    }
+
+    /**
+     * Get a instance of \DateTime with the current data time including milliseconds.
+     *
+     * @return \DateTime
+     */
+    private function currentDateTime()
+    {
+        $dateTime = \DateTime::createFromFormat('U.u', sprintf('%.6F', microtime(true)));
+        $dateTime->setTimezone(new \DateTimeZone(date_default_timezone_get()));
+
+        return $dateTime;
     }
 }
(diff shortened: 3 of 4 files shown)


boekkooi:
The 5.* failures are due to `Could not authenticate against github.com` I'm not sure why PHP 7 fails but it seems unrelated.

src/Model/Timestampable/TimestampableMethods.php
```diff
@@ -45,6 +47,7 @@ public function getUpdatedAt()
     public function setCreatedAt(\DateTime $createdAt)
     {
         $this->createdAt = $createdAt;
+        var_dump($this->createdAt->format('U.u'), 'SET');
```
kozmenkoav: Please do not forget to remove this line.

boekkooi: Removed .... and well spotted totally missed that one. Thanks!

src/Model/Timestampable/TimestampableMethods.php
```diff
@@ -10,6 +10,8 @@
  */
 
 namespace Knp\DoctrineBehaviors\Model\Timestampable;
+use Doctrine\ORM\Event\LifecycleEventArgs;
+use Doctrine\ORM\Event\PreUpdateEventArgs;
```
docteurklein: why those 2 use statements?

docteurklein:
Wouldn' t this have consequences on the doctrine mapping? can we store this format in a datetime column?

boekkooi:
@docteurklein Actually the current behaviour is suboptimal for some database. For example [Microsoft SQL Server 2008 and up](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Platforms/SQLServer2008Platform.php#L67) and [SAP Sybase SQL Anywhere 12 and up](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Platforms/SQLAnywhere12Platform.php#L58) uses microseconds. 

For other implementations there is no change expect when using a custom DateTimeType. This is something that people with MYSQL 5.6 and up can do by changing the `getSQLDeclaration` to use `DateTime(6)`.

P.S. Removed the unneeded use statements.

boekkooi:
Hey @docteurklein any chance of getting this merged?

docteurklein:
I'm still not sure if I like this to be enabled by default.
Wouldn't it better if we kept the current behavior but provide enogh extension points so that you can implement it just by configuring it the way you want?

boekkooi:
I'm not sure in what way a good extension point can be build and I also feel that it is not needed.

Let's take a look at why.

*Timestampable*
The current behaviours is to use a datetime without microseconds. For some databases at mentioned before this mean that the timestamp is inaccurate. (This is in my opinion a bug)
Within commit 2d0edf751ea2ad26b6e412ca6fbb14545f144193 the microseconds are added to the datetime. This ensures that for databases that support it the microseconds will be saved by doctrine, for databases that don't support it doctrine will automatically strip the microseconds so nothing changes.
  
*SoftDeletable*
As with Timestampable the datetime currently used is without microseconds and this again means that in some databases the timestamp is inaccurate.
With commit 4d32d6fe68aae2ba05d5dd93e47ae7599cecb808 the microseconds are added to the `deletedAt` datetime. This again will cause no changes for databases that don't support  microseconds because doctrine will format the datatime without it but it will allow databases with microseconds support to be more accurate.

*Loggable*
The current loggable behaviour stores the datetime without microseconds.
With commit 58bd792db9e49418c27eb1669af81db9914ea263 this behaviour is changed to always store the microseconds this is actually a bigger change because loggable will now always have microseconds for a datetime. In my opinion this is only better because with this change finding out what was logged at a specific time is more accurate and could help solve some bugs and make sorting a log easier. (as a side note a check could be added to only render the microseconds when available check `if ($datetime->format('u') != 0)`)

*General*
In general this PR has no effect other then giving doctrine and your log a more accurate datetime to work with.

- The doctrine/dbal [DateTimeType](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Types/DateTimeType.php#L57) uses the platforms `getDateTimeFormatString` which may use microseconds.
- This would never have been a discussion when PHP would always store microseconds for datetime now. This is PR is just a work around (related to https://bugs.php.net/bug.php?id=51950, https://bugs.php.net/bug.php?id=52519, https://bugs.php.net/bug.php?id=68506)

@docteurklein Could you please explain why this should not be enabled on default? or what other misgivings you have?

docteurklein:
I must admin the main reason is that I never saw microseconds persisted whatsoever.

I didn't know either that doctrine would gracefully handle the diff and thus that it was a BC break.
Now you bring this info to the table, I'm totally OK to reconsider my hesitation.

That said, I still think having an open/closed impl is better than hardcoding it (which is still the case before and after this PR).
For example, we could let the subscriber do the formatting, and add a `format` parameter to this subscriber .

But again, if doctrine handles it gracefully, I guess it's OK as is.

I'd still like to see if other people have feedback on this before merging it.
It'd be great if you added regression tests for this @boekkooi before going any further.

boekkooi:
@docteurklein Would it be better if I split this PR into 2 or 3 (Timestampable, SoftDeletable, Loggable)?

Since the formatting/database conversion is done by doctrine we don't need to add any more tests for Timestampable and SoftDeletable. Because we only provide doctrine with more accurate data.

For Loggable I do see your point that providing `dateTimeFormat` would be great and default that to using the doctrine platform [`getDateTimeFormatString`](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php#L3277).

@docteurklein I you think any more tests are needed could you please outline them for me because I don't see what I missed.

docteurklein:
lit's keep it like that. 
I didn't see you added tests before, so nevermind :).

Let's just see if anyone has feedback and merge.

boekkooi:
:+1:

boekkooi:
@docteurklein Any chance this PR can get merged?

docteurklein:
thanks :)

boekkooi:
No problem! Thank you for merging and the maintaining this package!

PR #288, opened 2016-02-05
TranslatableMethods: use late static bindings

meyerbaptiste:
Use late static bindings (replace `self::` by `static::`) for `TranslatableMethods::doTranslate()`.

Use case:
```php
abstract class Foo
{
    use Translatable;
}

class Baz
{
    use Translation;

    public static function getTranslatableEntityClass()
    {
        return Bar::class;
    }
}

class Bar extends Foo
{
    public static function getTranslationEntityClass()
    {
        return Baz::class;
    }
}

### DIFF
--- a/src/Model/Translatable/TranslatableMethods.php
+++ b/src/Model/Translatable/TranslatableMethods.php
@@ -114,7 +114,7 @@ trait TranslatableMethods
             }
         }
 
-        $class       = self::getTranslationEntityClass();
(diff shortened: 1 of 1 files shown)


nlemahieu:
:+1:

K-mos:
:+1:

meyerbaptiste:
Travis failed because of Composer.

docteurklein:
thanks!

Issue #297, opened 2016-03-09
Translation stopped working in Symfony 3.0.3 project with Doctrine behaviors 1.3


yvesblancquaert:
Hi,

I've successfully used translatable in previous projects based on Symfony 2.8 and Doctrine Behaviors 1.2. In my current Symfony 3.0.3 project where I use Doctrine Behaviors 1.3 I keep getting the following error message: 
[Doctrine\ORM\Mapping\MappingException] No identifier/primary key specified for Entity "AppBundle\Entity\CanvasAclTranslation". Every Entity must have an identifier/primary key.

I've added the bundle to the AppKernel and only activated translatable via the config.yml file.
I've attached an overview of current installed packages and the entity classes.

Regards
Yves

[CanvasAcl.txt](https://github.com/KnpLabs/DoctrineBehaviors/files/164796/CanvasAcl.txt)
[CanvasAclTranslation.txt](https://github.com/KnpLabs/DoctrineBehaviors/files/164797/CanvasAclTranslation.txt)
[InstalledPackages.txt](https://github.com/KnpLabs/DoctrineBehaviors/files/164798/InstalledPackages.txt)

NiR-:
Hey @yvesblancquaert, I looked at your CanvasAclTranslation and it appears that you are using the `Translatable` trait in the translation. You should try to fix that, and I think all your problems will disappear :)

yvesblancquaert:
Hey @NiR- Ouch, that hurts :-) Thank you very much for looking at it. Copy-paste of class code was not that smart. All my problems have disappeared.

NiR-:
Good news :)

Issue #292, opened 2016-02-25
createdAt and updatedAt are missing


CharlyPoppins:
Hi,

I don't don't what happened, I just found out that createdAt and updatedAt was missing in my database. deletedAt is still in my database.

I've had this in config.yml and it solve the problem.
knp_doctrine_behaviors:
    soft_deletable: true
    timestampable: true

Issue #286, opened 2016-01-27
Default configuration doesn't work properly


kozmenkoav:
Subscribers are not enabled by default, but documentation says (https://github.com/KnpLabs/DoctrineBehaviors/blame/1.3.0/README.md#L32) that they should be.
I think the issue is near this line https://github.com/KnpLabs/DoctrineBehaviors/blob/1.3.0/src/Bundle/DependencyInjection/Configuration.php#L15

My translations disappeared until I added lines:
```yml
knp_doctrine_behaviors:
    translatable: true
```

kozmenkoav:
thanks!

Issue #281, opened 2016-01-14
[BC Break] Translation stopped working after 1.2.0 -> 1.3.0 update on Doctrine 2.6


kiler129:
Hi!
I think I maybe found a bug in the newest 1.3.0 release. I'm using `translatable` from `DoctrineBehaviors` along with [a2lix/translation-form-bundle](https://github.com/a2lix/TranslationFormBundle) & [Sonata](https://github.com/sonata-project/SonataAdminBundle).
After upgrade I noticed all translations are missing. To pinpoint that issue I downgraded Doctrine from 2.6.1 to 2.6.0 - result was the same. Next I've tried downgrading `DoctrineBehaviors` to 1.2.0, and bingo - everything started working.
I wrote minimal code to test it:
```php
$pageRepo = $this->getDoctrine()->getRepository('XyzBundle:Page');
$page = $pageRepo->find(1);
var_dump($page);
```
On 1.3.0 it returns nulls where on 1.2.0 proper vales are returned.

Here're my implementation details (simplified to single-field, no setters, no comments to save space):
**Mappings**
```xml
<?xml version="1.0" encoding="utf-8"?>
<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                  xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping http://doctrine-project.org/schemas/orm/doctrine-mapping.xsd">
    <entity name="FooCompany\XyzBundle\Entity\Page" table="pages">
        <id name="id" type="integer" column="id">
            <generator strategy="AUTO"/>
            <options>
                <option name="unsigned">true</option>
            </options>
        </id>
        <field name="userEditable" type="boolean" column="user_editable" nullable="false"/>
    </entity>
</doctrine-mapping>
```
```xml
<?xml version="1.0" encoding="utf-8"?>
<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                  xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping http://doctrine-project.org/schemas/orm/doctrine-mapping.xsd">
    <entity name="FooCompany\XyzBundle\Entity\PageTranslation" table="page_translations">
[...]

PR #298, opened 2016-03-09
Fix #267: if no config specified all behaviors are enabled

### LINKED ISSUE #267: Commit a52a92a disables all subscribers
According to the updated documentation, commit a52a92a intended to make all configurations enabled by default when the root node `knp_doctrine_behaviors` is empty. However, when tested on Symfony 2.7.6, the actual behavior is that all behaviors are disabled by default. Looks like the `treatNullLike([... => true ])` in `Configuration` class has no effect. Can anyone verify this behavior? Thanks.

NiR-:


docteurklein:
fixes #297, #292, #286, #281, #272, #270

yvesblancquaert:
Error after update to 1.3.1
capture

yvesblancquaert:
v1.3.2 didn't solve issue #297 
capture1

docteurklein:
@yvesblancquaert thanks for the quick feedback!
I guess we need to figure out what's the problem first, before tagging anything :)

Could you check if the TranslatableSubscriber class is even correctly called?
It seems that it does not automagically map the IDs anymore (https://github.com/KnpLabs/DoctrineBehaviors/blob/master/src/ORM/Translatable/TranslatableSubscriber.php#L93).

Can you debug this part and look if `mapId` is called?

PR #299, opened 2016-03-09
Update Configuration.php

lmerotta:
Remove unused variable that causes composer errors, regarding PR #298


NiR-:
I've fix the notice error in #300, and a new release (1.3.2) is available. Thanks for the PR @lmerotta.

PR #303, opened 2016-03-18
Updated master branch alias

boekkooi:
Hi there,

This is just a plain simple dump of the master branch alias so people using master because a PR was merged but not yet released can use something like ~1.3, >= 1.3.3 and it will pick up the master branch.


Einenlum:
Outdated PR now, so closed. Thanks for your time.

PR #306, opened 2016-03-18
[Sortable] Add ability to discard reordering

bocharsky-bw:
This PR do **not** break BC.

### DIFF
--- a/src/Model/Sortable/Sortable.php
+++ b/src/Model/Sortable/Sortable.php
@@ -39,12 +39,26 @@ trait Sortable
     }
 
     /**
+     * @param bool $reordered.
+     * 
      * @return $this
      */
-    public function setReordered()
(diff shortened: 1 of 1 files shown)


Einenlum:
This looks ok to me :). Does that make sense to you @docteurklein ?

src/Model/Sortable/Sortable.php
```diff
@@ -39,12 +39,26 @@ public function isReordered()
...
     }
+
+    /**
+     * Reorder items
+     * 
+     * @return $this
+     */
+    public function reorder()
+    {
+        $this->reordered = true;
```
Einenlum: Then maybe just:

```
return $this->setReordered(true);
```

TomasVotruba:
Closing as outdated

PR #315, opened 2016-05-31
[Translatable] Check if entity is really translatable before setting the locales

### LINKED ISSUE #314: [Translateable] Attempted to call an undefined method named "setDefaultLocale" of class ...
hi,
i use this with sylius, but i got errors on many entities which doesn't have the property locale on the main entity but got it in an abstract class; there is a conflict.

```
Attempted to call an undefined method named "setDefaultLocale" of class "Proxies\__CG__\Sylius\Component\Core\Model\Taxon".
```

How to disabled the TranslatableSubscriber for these entities ?

pjedrzejewski:
Fixes #314.

### DIFF
--- a/src/ORM/Translatable/TranslatableSubscriber.php
+++ b/src/ORM/Translatable/TranslatableSubscriber.php
@@ -314,6 +314,10 @@ class TranslatableSubscriber extends AbstractSubscriber
         $entity        = $eventArgs->getEntity();
         $classMetadata = $em->getClassMetadata(get_class($entity));
 
+        if (!$this->isTranslatable($entity)) {
+            return;
+        }
+
         if (!$this->getClassAnalyzer()->hasMethod($classMetadata->reflClass, 'setCurrentLocale')) {
             return;
         }


src/ORM/Translatable/TranslatableSubscriber.php
```diff
@@ -314,6 +314,10 @@ private function setLocales(LifecycleEventArgs $eventArgs)
         $entity        = $eventArgs->getEntity();
         $classMetadata = $em->getClassMetadata(get_class($entity));
 
+        if (!$this->isTranslatable($entity)) {
+            return;
+        }
+
         if (!$this->getClassAnalyzer()->hasMethod($classMetadata->reflClass, 'setCurrentLocale')) {
```
lemoinem: I'd think you can remove this `if`... If you check if the `$entity` is Translatable, the method's availability should be implied from that...

marcoslibre:
@pjedrzejewski Why did you close your PR ?
I think **it is** relevant to check if the entity is translatable before setting the locale.
If an entity is not translatable and has a setCurrentLocale method, but not a setDefaultLocale it can crash !
You should re-open this PR.

lemoinem:
@pjedrzejewski If you closed the PR because of my comment. I was referring to the second `if`, the one the line comment is on. Not the one you added. I, also, think this check is very good idea.

pjedrzejewski:
I will try to finish it.

tamert:
The form's view data is expected to be an instance of class Proxies\__CG__\App\Entity\UserLevelTranslation, but is an instance of class App\Entity\UserLevelTranslation. You can avoid this error by setting the "data_class" option to null or by adding a view transformer that transforms an instance of class App\Entity\UserLevelTranslation to an instance of Proxies\__CG__\App\Entity\UserLevelTranslation.

I dont understont ? only one record add "Proxies\__CG__"

PR #316, opened 2016-06-02
Fixes for Symfony 3.1

tarlepp:
Not quoting a scalar starting with the "%" indicator character is deprecated since Symfony 3.1 and will throw a ParseException in 4.0.

https://github.com/symfony/symfony/blob/3.1/UPGRADE-3.1.md#yaml


MaZderMind:
nice catch. release 1.3.2 fails with current PHPUnit 5.4.7, because it ships a newer Yml-Parser which shadows symfony's own impl and fails with the above error.

Is there a plan for an 1.4 release? 1.3.2 is nearly 2 months old by now.
Thanks for the awsome work!

PR #317, opened 2016-06-13
Fix interaction between translations & joined inheritance

lemoinem:
When using a JoinedInheritance for translations, the `TranslatableSubscriber` will try to map the children classes, which will create a new unique index on a column existing only in the parent class' table (`translatable_id`).

This crashes Doctrine.

Anyway, the fields need, associations and indexes only to be mapped for the root class of the translation hierarchy, since they will all be inherited. This is true for every kind of inheritance and every behaviour.

This PR fixes the issue for Translatable.

### DIFF
--- a/src/ORM/Translatable/TranslatableSubscriber.php
+++ b/src/ORM/Translatable/TranslatableSubscriber.php
@@ -41,11 +41,11 @@ class TranslatableSubscriber extends AbstractSubscriber
     private $translatableFetchMode;
     private $translationFetchMode;
 
-    public function __construct(ClassAnalyzer $classAnalyzer, $isRecursive, callable $currentLocaleCallable = null,
+    public function __construct(ClassAnalyzer $classAnalyzer, callable $currentLocaleCallable = null,
                                 callable $defaultLocaleCallable = null,$translatableTrait, $translationTrait,
                                 $translatableFetchMode, $translationFetchMode)
     {
-        parent::__construct($classAnalyzer, $isRecursive);
+        parent::__construct($classAnalyzer, false);
 
         $this->currentLocaleCallable = $currentLocaleCallable;
         $this->defaultLocaleCallable = $defaultLocaleCallable;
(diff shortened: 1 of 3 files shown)


ksom:
Can you merge please ? Since June 13 it's open and no comment.
It's really stuck to develop with `inheritance_type` as `JOINED`.

tarlepp:
@ksom I'm not a collaborator of this repository.

ksom:
@tarlepp Thanks, I corrected it, sorry for having tagged you in.

src/ORM/Translatable/TranslatableSubscriber.php
```diff
@@ -68,11 +68,11 @@ public function loadClassMetadata(LoadClassMetadataEventArgs $eventArgs)
             return;
         }
 
-        if ($this->isTranslatable($classMetadata)) {
+        if ($this->isTranslatable($classMetadata, false)) {
```
docteurklein: why are you passing false here, instead of directly putting false below? (see next comment)

src/ORM/Translatable/TranslatableSubscriber.php
```diff
@@ -280,22 +280,26 @@ private function hasUniqueTranslationConstraint(ClassMetadata $classMetadata, $n
...
-     * @param bool          $isRecursive   true to check for parent classes until found
+     * @param bool|null     $isRecursive   true to check for parent classes until found
      *
      * @return boolean
      */
-    private function isTranslatable(ClassMetadata $classMetadata, $isRecursive = false)
+    private function isTranslatable(ClassMetadata $classMetadata, $isRecursive = null)
     {
-        return $this->getClassAnalyzer()->hasTrait($classMetadata->reflClass, $this->translatableTrait, $this->isRecursive);
+        return $this->getClassAnalyzer()->hasTrait($classMetadata->reflClass, $this->translatableTrait, null !== $isRecursive ? $isRecursive : $this->isRecursive);
```
docteurklein: just put false here

docteurklein:
I think you overcomplicated the impl, unless I'm missing something?

BTW, As soon as you have complicated (inheritance-mapping), it might be easier to just make the mapping by yourself.

Another point to take into consideration: there are PRs out there that will remove this trait detection system and only use interfaces.
It won't remove the dumb auto mapping tho.

docteurklein:
FYI, I think we recently tried translations on an inheritance tree with JOINED strategy and it just seed to work.

lemoinem:
@docteurklein I simply disabled the recursivity for some calls in the detection of the trait. I don't think that's overcomplicating it... The only was I see of simplifying it further would be to remove and disable the recursive flag in the ClassAnalyzer all across the repository, but I'm not sure what this would mean for the rest of the Behaviors...

BTW, I am aware of the other PRs, I prepared some of them too, and made sure the issue would not repeat itself with the new process (and in order not to report it, we still have to rely on the traits for the "dump auto-mapping" ;) ).

lemoinem:
@docteurklein What about this version, since there is no dependence on the $isRecursive flag, I removed it entirely...

docteurklein:
perfect! that's what I thought: the $isRecursive is local (internal) to this subscriber and won't change anything to the other subscribers.
Let's try that.

PR #318, opened 2016-06-16
[Translatable] Remove Translation entity when empty

Laulibrius:
Instead of updating a translation with empty data, why not removing this translation according to the return value of isEmpty() method ?


TomasVotruba:
Closing as outdated

Issue #241, opened 2015-07-29
Use interfaces instead of traits


bocharsky-bw:
With interfaces implementation code looks more clearer, readable and correct than with current trait implementation. Developers could easily override properties and methods with their customs. Also we get auto-completion by IDEs in code. More discusion was in PR https://github.com/KnpLabs/DoctrineBehaviors/pull/237 .

There are a list of traits that should be refactored with interfaces:
- [x] - tree (in https://github.com/KnpLabs/DoctrineBehaviors/pull/246)
- [x] - translatable (in #322)
- [x] - timestampable (in https://github.com/KnpLabs/DoctrineBehaviors/pull/242)
- [x] - softDeletable (in https://github.com/KnpLabs/DoctrineBehaviors/pull/248)
- [x] - blameable (in #323)
- [x] - loggable (in #324)
- [x] - geocodable (in #325)
- [x] - filterable
- [x] - sluggable (in https://github.com/KnpLabs/DoctrineBehaviors/pull/237)
- [x] - sortable (in https://github.com/KnpLabs/DoctrineBehaviors/pull/249)

migmolrod:
Hum. Maybe (very probably) I'm missing something, but with that approach users would have to implement the very same methods over and over in each entity using the behavior, right?

The good thing about 'traits' is they are so 'DRY'... (Don't Repeat Yourself). Wouldn't we be losing that with interfaces?

I can't see anything obscure or incorrect about traits, at all. They are just tools and I think they fit very well in this case.
About the auto-completion, I think there are ways to help with that, using ```@property``` PHPDoc tag (never used it myself yet, but I've read about it somewhere I don't remember now)

bocharsky-bw:
@migmolrod Well, look closely to this PRs and changes that I made in them. I don't propose to remove traits that are awesome for DRY. I only suggest to add interfaces to this behaviors and do some refactor for using them.

So we keep all traits and provide a much clearer arvhitecture with interfaces as I explain in https://github.com/KnpLabs/DoctrineBehaviors/pull/237#issuecomment-122708798 .

migmolrod:
Ahhh. Sorry for the misconception then.

bocharsky-bw:
@migmolrod no problem. Seems these PRs maintainers review a bit hang up.

ruscon:
+1

NiR-:
Also using interfaces rather than reflection on properties will prevent any conflict. :+1:

[...]

PR #322, opened 2016-09-27
[Translatable] Rewrite Translatable/Translation to use Interfaces instead of Traits

lemoinem:
Here is the Translatable part for #241

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ same folder of Category entity by default.
 The default naming convention (or its customization via trait methods) avoids you to manually handle entity associations.
 It is handled automatically by the TranslationSubscriber.
 
-In order to use the Translatable trait, you will have to create this `CategoryTranslation` entity.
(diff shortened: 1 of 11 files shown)


lemoinem:
I'm not entirely sure why the travis build failed, but it seems to have the same issue on master already, so... I'm gonna guess it's unrelated to this PR.

Pierstoval:
```
s\Knp\DoctrineBehaviors\ORM\Timestampable
 [x] It should initialize create and update datetime when created
 [ ] It should modify update datetime when updated but not the creation datetime
 [ ] It should return the same datetime when not updated
 [ ] It should modify update datetime only once
```

Comes from timestampable ;)

lemoinem:
@Pierstoval Oh, ok. I have the same fails in local... The errors are harder to spot with this output... (Which is new to me, too...) but now I know what to look for, it will easier. Thanks!

docteurklein:
BTW, I think once we moved to interfaces, we can get rid of class analyzer. that would be so good

PR #323, opened 2016-09-27
[Blameable] Rewrite Blameable to use Interfaces instead of Traits

lemoinem:
Here is the Blameable part for #241


TomasVotruba:
Implemented in https://github.com/KnpLabs/DoctrineBehaviors/pull/442

TomasVotruba:
Thank you :heart:

PR #329, opened 2016-09-28
Add Unique Index name generator

wucdbm:
In an edge case I had, the name of the unique index for the translations table was too long and MySQL refused it. I needed a way of getting this done quickly, so I implemented an interface & a default implementation that does (or should not) change the behavior in any way, but allows for custom implementations.

What you need to do in order to take advantage of it is to either replace the class parameter or the service altogether, `config/orm-services.yml` for reference.


TomasVotruba:
Closing as outdated

PR #332, opened 2016-10-06
Fix: isEmpty() return true if it's empty

corentinheadoo:
The function isEmpty() return always false... fix it


src/Model/Translatable/TranslationMethods.php
```diff
@@ -93,6 +93,13 @@ public function getLocale()
      */
     public function isEmpty()
     {
-        return false;
+        foreach (get_object_vars($this) as $var => $value) {
+            if (in_array($var, array('id', 'translatable', 'locale')))
```
cursedcoder: please use short syntax array

src/Model/Translatable/TranslationMethods.php
```diff
@@ -93,6 +93,13 @@ public function getLocale()
...
+                continue;
+            
+            if (!empty($value))
+                return false;
```
cursedcoder: please use enclosing brackets `{}`

src/Model/Translatable/TranslationMethods.php
```diff
@@ -93,6 +93,13 @@ public function getLocale()
...
+        }
+        return true;
```
cursedcoder: return statement should be separated by 1 space line

Pierstoval:
LGTM 👍 /cc @cursedcoder

docteurklein:
this method is meant to be overriden. e just provide a dummy impl to avoid extra work.
do we want a *smart* isEmpty?

PR #350, opened 2017-05-15
Fix disabling softdeletable

ossinkine:
In `orm-services.yml` the service defined as `knp.doctrine_behaviors.softdeletable_subscriber` but `DoctrineBehaviorsExtension` tries to remove `knp.doctrine_behaviors.soft_deletable_subscriber`


NiR-:
Thank you @ossinkine for spotting this issue :+1:

PR #364, opened 2017-09-25
[Tree] Fix flat tree query

### LINKED ISSUE #357: [Tree] Flat tree query issue
The `getFlatTreeQB()` method is using only LIKE operator to compare paths and can return wrong results.

Lets asume we use IDs for path generation. When getting flat tree for entity with id === 1, the query is something like `WHERE materialized_path LIKE '/1%'`. That means if i have element assigned to parent with id 10, 11, 12 and so on, it will be returned either.

Einenlum:
Fixes #357

### DIFF
--- a/src/ORM/Tree/Tree.php
+++ b/src/ORM/Tree/Tree.php
@@ -4,6 +4,7 @@ namespace Knp\DoctrineBehaviors\ORM\Tree;
 
 use Knp\DoctrineBehaviors\Model\Tree\NodeInterface;
 use Doctrine\ORM\QueryBuilder;
+use Knp\DoctrineBehaviors\Model\Tree\Node;
 
 trait Tree
 {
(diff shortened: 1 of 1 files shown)


src/ORM/Tree/Tree.php
```diff
@@ -50,19 +51,34 @@ public function getRootNodes($rootAlias = 't')
...
-    public function getTree($path = '', $rootAlias = 't')
-    {
-        $results = $this->getFlatTree($path, $rootAlias);
+    public function getTree(
+        $path = '',
+        $rootAlias = 't',
+        $materializedPathSeparator = null
+    ) {
+        $materializedPathSeparator = $materializedPathSeparator
+            ?? self::getMaterializedPathSeparator()
```
NiR-: * This repo still requires php >= 5.4.
* If the class using this trait has `getMaterializedPathSeparator()`, for what use case is `$materializedPathSeparator` useful?

src/ORM/Tree/Tree.php
```diff
@@ -50,19 +51,34 @@ public function getRootNodes($rootAlias = 't')
...
         return $this->buildTree($results);
     }
 
-    public function getTreeExceptNodeAndItsChildrenQB(NodeInterface $entity, $rootAlias = 't')
+    public function getTreeExceptNodeAndItsChildrenQB(NodeInterface $entity, $rootAlias = 't', $materializedPathSeparator = null)
     {
-        return $this->getFlatTreeQB('', $rootAlias)
-            ->andWhere($rootAlias.'.materializedPath NOT LIKE :except_path')
+        $materializedPathSeparator = $materializedPathSeparator
+            ?? self::getMaterializedPathSeparator()
```
NiR-: Same here

src/ORM/Tree/Tree.php
```diff
@@ -50,19 +51,34 @@ public function getRootNodes($rootAlias = 't')
...
+        ;
+
+        $qb = $this->getFlatTreeQB('', $rootAlias, $materializedPathSeparator);
+
+        return $qb
+            ->andWhere($rootAlias.'.materializedPath NOT LIKE :except_exact_path')
+            ->andWhere($rootAlias.'.materializedPath NOT LIKE :except_parent_path')
```
NiR-: You add two `WHERE NOT LIKE` clauses here but you use the QB coming from `getFlatTreeQB` which adds a `WHERE (LIKE ... OR LIKE ...)` clause. It's a bit strange imho.

src/ORM/Tree/Tree.php
```diff
@@ -94,12 +110,27 @@ public function buildTree($results)
...
-        $qb = $this->createQueryBuilder($rootAlias)
-            ->andWhere($rootAlias.'.materializedPath LIKE :path')
+        $materializedPathSeparator = $materializedPathSeparator
+            ?? self::getMaterializedPathSeparator()
+        ;
+
+        $qb = $this->createQueryBuilder($rootAlias);
+
+        $qb
+            ->andWhere(
```
NiR-: Maybe we can even avoid this clause when `$path` is an empty.

src/ORM/Tree/Tree.php
```diff
@@ -94,12 +110,27 @@ public function buildTree($results)
      *
      * @return QueryBuilder
      */
-    public function getFlatTreeQB($path = '', $rootAlias = 't')
+    public function getFlatTreeQB($path = '', $rootAlias = 't', $materializedPathSeparator = null)
     {
-        $qb = $this->createQueryBuilder($rootAlias)
-            ->andWhere($rootAlias.'.materializedPath LIKE :path')
+        $materializedPathSeparator = $materializedPathSeparator
```
NiR-: Same as previous comments about `$materializedPathSeperator`.

src/ORM/Tree/Tree.php
```diff
@@ -130,12 +161,26 @@ protected function addFlatTreeConditions(QueryBuilder $qb)
      *
      * @return array the flat resultset
      */
-    public function getFlatTree($path, $rootAlias = 't')
+    public function getFlatTree($path, $rootAlias = 't', $materializedPathSeparator = null)
     {
+        $materializedPathSeparator = $materializedPathSeparator
```
NiR-: Same again

TomasVotruba:
If this is ~~rebased~~ sent (there were so many changes) on current `master` + covered with test, I'd be happy to merge :+1:

PR #365, opened 2017-09-27
Fix drop php < 7

Einenlum:


NiR-:
So we drop support for PHP 7.0 too?

PR #367, opened 2017-09-27
Allow someone to extends a Traits

skrosoft:
Allow someone to extends a Traits.

Exemple:

```
<?php
// ...

use Knp\DoctrineBehaviors\Model\Tree\Node;

trait MySomethingable
{
    use Node;

    // ...
	
    public function myFunction(MyInterface $parent = null)
    {
        // ...

        $this->setParentNode($parent);

        // ...
    }
	
    // ...
}
```

:)

### DIFF
--- a/src/Reflection/ClassAnalyzer.php
+++ b/src/Reflection/ClassAnalyzer.php
@@ -26,6 +26,12 @@ class ClassAnalyzer
         if (in_array($traitName, $class->getTraitNames())) {
             return true;
         }
+        
+        foreach($class->getTraitNames() as $forTraitName){
+            if ($this->hasTrait(new \ReflectionClass($forTraitName), $traitName, true)){
+                return true;
+            }
+        }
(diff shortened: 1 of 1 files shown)


Einenlum:
Hi @skrosoft.

If you want to "extend" traits, you can always create a trait that uses another trait, like this answer: https://stackoverflow.com/a/37687295

skrosoft:
Hi @Einenlum . Yes I know how to “extend” traits. But without my code, if you extend a trait, the command “doctrine:scheme:update” will not create the new fields in the database. Please check my code and tell me if you need more information.

skrosoft:
Ps: in the case of “Tree/Node”, the field “materialized path” would not be created.

TomasVotruba:
Could you add test for it?

TomasVotruba:
This will be replaced by interfaces, see #241

PR #368, opened 2017-10-16
[Translatable] Fix #362 - Prevent duplicate translation entry

### LINKED ISSUE #362: [Translatable] Empty value cause problem in v1.4.1
I have entity:
```php
class ProductVariantTranslation
{
	use Translation;

	/**
	 * @ORM\Column(type="string", nullable=true)
	 * @var string
	 */
	protected $name;
}
```

The problem was caused, if i have empty name property and call this

```php
$variant->setName('');
$variant->mergeNewTranslations();
```

returns

```Doctrine\\DBAL\\Exception\\UniqueConstraintViolationException(code: 0): An exception occurred while executing 'INSERT INTO product_variant_translation (name, locale, translatable_id) VALUES (?, ?, ?)' with params [\"\", \"cs\", 1]:\```

because this function evaluates this translation as empty in version 1.4.1

https://github.com/KnpLabs/DoctrineBehaviors/blob/6b99da7e6bb5570657bec6c3d7b55455b270fa19/src/Model/Translatable/TranslationMethods.php#L94

StefanNiederkofler:
Even if a translation is empty, it should be returned if no fallback was found. Creating a new one, cause a duplicate entry.

### DIFF
--- a/src/Model/Translatable/TranslatableMethods.php
+++ b/src/Model/Translatable/TranslatableMethods.php
@@ -114,6 +114,10 @@ trait TranslatableMethods
             }
         }
 
+        if ($translation) {
+            return $translation;
+        }
+
         $class       = static::getTranslationEntityClass();
         $translation = new $class();
         $translation->setLocale($locale);


skududufru:
Please accepte this one. A lot a people have issues cause to that.

docteurklein:
Wouldn't overriding `isEmpty` resolve the problem? (see line 102)

StefanNiederkofler:
@docteurklein would be possible, but in fact isEmpty does what it should. The entity IS empty, but exists nevertheless. Empty and existing is not the same.

diaborn19:
ping @corentinheadoo

corentinheadoo:
@diaborn19 
I have overridden `isEmpty` for another bug, to get fallback on existing translation with all empty fields.  
With this `isEmpty` fix, the duplicate entry bug appears

Soullivaneuh:
Could you please add a proof test?

NiR-:
I believe empty translations should not be persisted in the first place. IMHO you should wait the next release and then clean up your database.

DessaugesAntoine:
I have the same problem and can't wait the pull request to be merge. So here a solution :
Create a patch you can apply with composer.
See the diff in the pull request and create two file, the original package file and the same file with the fix. Create a .diff file with git and store it in your project.
You can now apply this .diff file with composer.
Here some detailled informations : https://punkt.de/en/blog/2017/patches-with-composer.html

Hope this can help pepole (I lost so many time to find where the problem was and how to find a solution...)

PR #376, opened 2018-01-21
Add "Uuidable" trait based on Ramsey UUID4

petegore:


Soullivaneuh:
Could you please take a look at Travis and fix your commits history?

NiR-:
@petegore Good idea! There's been an architecture overhaul started times ago and still not finished at this time. It would merely get rid of the messy trait detection and instead use proper interfaces to detect when mapping is necessary. I think it would be cool if your PR could get this right from the beginning.

TomasVotruba:
Pretty neat feature! I'll integrate it into next version, after the clean up

Thank you for the PR

TomasVotruba:
Thank you for the feature! :+1: 

Rebased on current master in https://github.com/KnpLabs/DoctrineBehaviors/pull/450
and implemented

PR #382, opened 2018-03-23
[Translatable] Do not persist new translations if empty

giuliapellegrini:
Do not persist new translations if empty.


NiR-:
Thank you @giuliapellegrini :+1:

PR #387, opened 2018-05-02
Rewrite Translatable/Translation to use Interfaces instead of Traits

NiR-:
This is PR #322 rebased on current master branch and with minor fixes. Credits go to @lemoinem 

Left to do:
* [ ] Update readme and upgrade guide
* [ ] Implement a compatibility layer to smooth transition from trait to interface (in a separate PR)

### DIFF
--- a/src/Model/Translatable/TranslatableMethods.php
+++ b/src/Model/Translatable/TranslatableMethods.php
@@ -43,11 +43,11 @@ trait TranslatableMethods
     /**
      * Adds new translation.
      *
-     * @param Translation $translation The translation
+     * @param TranslationInterface $translation The translation
      *
      * @return $this
      */
-    public function addTranslation($translation)
+    public function addTranslation(TranslationInterface $translation)
     {
         $this->getTranslations()->set((string)$translation->getLocale(), $translation);
         $translation->setTranslatable($this);
@@ -58,9 +58,9 @@ trait TranslatableMethods
     /**
      * Removes specific translation.
      *
-     * @param Translation $translation The translation
+     * @param TranslationInterface $translation The translation
      */
-    public function removeTranslation($translation)
+    public function removeTranslation(TranslationInterface $translation)
     {
         $this->getTranslations()->removeElement($translation);
     }
@@ -74,7 +74,7 @@ trait TranslatableMethods
      * @param string $locale The locale (en, ru, fr) | null If null, will try with current locale
      * @param bool $fallbackToDefault Whether fallback to default locale
      *
-     * @return Translation
+     * @return TranslationInterface
      */
     public function translate($locale = null, $fallbackToDefault = true)
     {
@@ -90,7 +90,7 @@ trait TranslatableMethods
      * @param string $locale The locale (en, ru, fr) | null If null, will try with current locale
      * @param bool $fallbackToDefault Whether fallback to default locale
      *
-     * @return Translation
+     * @return TranslationInterface
      */
     protected function doTranslate($locale = null, $fallbackToDefault = true)
     {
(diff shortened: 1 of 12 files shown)


lemoinem:
Thanks @NiR- for the follow up and the credits. Sorry I didn't have the time to do the follow up.
I'm happy to this move forward and getting merged!

NiR-:
No worries, actually we were not really active on this repository in the previous months, too.

src/Model/Translatable/TranslatableMethods.php
```diff
@@ -43,11 +43,11 @@ public function getNewTranslations()
     /**
      * Adds new translation.
      *
-     * @param Translation $translation The translation
+     * @param TranslationsInterface $translation The translation
```
nicolasmure: Should be `TranslationInterface` (singular)

tests/Knp/DoctrineBehaviors/ORM/TranslatableTest.php
```diff
@@ -34,8 +34,8 @@ function()
             {
                 return 'en';
             },
-            'Knp\DoctrineBehaviors\Model\Translatable\Translatable',
-            'Knp\DoctrineBehaviors\Model\Translatable\Translation',
+            array('Knp\DoctrineBehaviors\Model\Translatable\Translatable', 'Knp\DoctrineBehaviors\Model\Translatable\TranslatableProperties'),
+            array('Knp\DoctrineBehaviors\Model\Translatable\Translation','Knp\DoctrineBehaviors\Model\Translatable\TranslationProperties',),
```
nicolasmure: Maybe it could be good to also have a test in case where a non array value (eg an object) is passed here instead of an array ?

src/Model/Translatable/TranslationInterface.php
```diff
@@ -0,0 +1,73 @@
...
+    public function getId();
+
+    /**
+     * Sets entity, that this translation should be mapped to.
+     *
+     * @param TranslatableInterface $translatable The translatable
+     *
+     * @return $this
+     */
+    public function setTranslatable($translatable);
```
nicolasmure: Can you typehint the parameter ?

src/ORM/Translatable/TranslatableSubscriber.php
```diff
@@ -68,11 +68,11 @@ public function loadClassMetadata(LoadClassMetadataEventArgs $eventArgs)
             return;
         }
 
-        if ($this->isTranslatable($classMetadata)) {
+        if ($this->requiresTranslatableMapping($classMetadata)) {
```
nicolasmure: The previous `isTranslatable` name was more understandable IMHO, and also in a way more accurate as `requires` could make think of a state awareness. WDYT?

src/ORM/Translatable/TranslatableSubscriber.php
```diff
@@ -68,11 +68,11 @@ public function loadClassMetadata(LoadClassMetadataEventArgs $eventArgs)
...
             $this->mapTranslatable($classMetadata);
         }
 
-        if ($this->isTranslation($classMetadata)) {
+        if ($this->requiresTranslationMapping($classMetadata)) {
```
nicolasmure: same here about the name

tests/Knp/DoctrineBehaviors/ORM/TranslatableTest.php
NiR-: Right, and it makes me think that we should provide a compatibility layer to not break BC.

src/Model/Translatable/TranslationInterface.php
NiR-: I added it but I'm not sure if I should add other missing types (notably return types) in this PR or create a dedicated one, mostly because I would have to modify traits (I wouldn't make the diff cluttered). WDYT?

nicolasmure: Thanks, yes I agree that return typehints should be added in a separate PR :)

src/Model/Translatable/TranslatableMethods.php
AlucardleVash: :+1:

TomasVotruba:
Covered by #442

Issue #388, opened 2018-05-03
[Translatable] Shrink locale column


NiR-:
During a database migration from `utf8` charset to `utf8mb4`, I realized that `locale` columns coming from this library are `VARCHAR(255)`. This works, but it's not optimum. It would be better to shrink down these columns to 2 chars.

Unfortunately, this would be a BC break and thus it could be done only for the next major release.

CoalaJoe:
What about en-us and en-gb? Maybe 2 is too small.

NiR-:
@CoalaJoe Indeed, I checked Symfony routing to see what is allowed, and determined 2 would be ok for this library; But as it can be used without Symfony, so limiting to 5 seems more reasonable :)

sylvaincombes:
imho we should have an upper limit and let the maxlength possible (190 if we want utf8mb4 compliance)

For example this is a valid locale with 3 components : zh_Hant_CN (Language code, Script code, Country code), but this one is a valid too de_DE_u_co_phonebk_cu_ddm

As seen on the CLDR site we can have :

```
unicode_language_subtag(2)_unicode_script_subtag(4)_unicode_region_subtag(2)_unicode_variant_subtag(8?)_[key_type, key_type, ...]
```
On real projects the longest I have used is zh_Hant_CN / zh_Hans_CN but I think it's a better strategy to give more room for developers choice and flexibility

References :

- https://www.unicode.org/reports/tr35/tr35-17.html#Locale
- https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

TomasVotruba:
Resolved by https://github.com/KnpLabs/DoctrineBehaviors/pull/389

PR #389, opened 2018-05-08
Shrink locale columns to 5 chars

### LINKED ISSUE #384: SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was t     oo long; max key length is 767 bytes
Hi,

I'm getting this error:

 

> SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was t  
>   oo long; max key length is 767 bytes  

How can I change default lenght of locale columnt(it should be 190)?

NiR-:
Currently, the mapping for locale columns does not provide any length,
thus Doctrine falls back to the default size: 255 chars.

Not only such size is useless, but it also prevents using this library
with utf8mb4 charset, because InnoDB indexes are capped at 767 bytes,
and unfortunately: 4 * 255 = 1020 bytes.

For reference, here is the error with utf8mb4 charset:
> SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key
> was too long; max key length is 767 bytes

Fixes #384 and #388.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,9 @@
+# 2.0.0
+
+Translatable:
+* `locale` columns are now capped at 5 chars. You might need to manually define
+  `locale` mapping, if you store anything lengthier.
+
 # 1.0.2 to 1.0.x-dev
 
 Most occurences of "listener" have been replaced with "subscriber" to honor a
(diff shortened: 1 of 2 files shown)


NiR-:
Following comment on the initial issue (#388), I increased the limit to 5 chars.

TomasVotruba:
Thank you for PR and nice detailed explanation :+1:

PR #390, opened 2018-05-09
Document master and v1 branches

NiR-:
IMHO we need to prepare a `v2` for upcoming changes breaking BC (for instance #241).

I didn't create `v1` branch so far, I'd like some feedback first. I think we need to either:
* Create a `v2` branch, keep v1 on `master` for now and put v2 on `master` later, when it got released ;
* Or create `v1` right now and put `v2` on master, but we have to change the base branch of every current PR.

@nicolasmure @Soullivaneuh WDYT?

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,10 @@
 
 [![Build Status](https://travis-ci.org/KnpLabs/DoctrineBehaviors.svg?branch=master)](http://travis-ci.org/KnpLabs/DoctrineBehaviors)
 
+* Branch `master` contains the next v2.0
+* Branch `v1` contains the current v1.x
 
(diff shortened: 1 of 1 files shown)


nicolasmure:
IMHO `master` should reflect the most advanced state of the project, so I think `master` should point to `v2` as soon as `v2` is created.
That being said, there are a lot of open PRs so I see the point of option `#1`. Are there some PR which aren't relevant anymore, and thus can be closed ? Do you have an idea for how long the `v2` branch will exist before being released ? Is the #241 refacto the 'only' changes to bring to the v2 ? If yes, I think we could go for option `#2` as the changes are looking ready enough to be released (from the quick look I gave at it, correct me if I'm wrong).

PR #392, opened 2018-05-24
Only set locales on entities managed by knp translations

jordisala1991:
There are cases where you could have a entity with the setCurrentLocale method, but that is not a translatable entity managed by KNP.

This PR only sets the default and current locales on entities that are translatable instead of relying on a method existance.

Closes https://github.com/KnpLabs/DoctrineBehaviors/issues/314
Closes https://github.com/KnpLabs/DoctrineBehaviors/pull/315

It also fixes compatibility with Sylius, because it uses its own translations implementation, and it is incompatible with having this other bundle installed.

### DIFF
--- a/src/ORM/Translatable/TranslatableSubscriber.php
+++ b/src/ORM/Translatable/TranslatableSubscriber.php
@@ -316,7 +316,7 @@ class TranslatableSubscriber extends AbstractSubscriber
         $entity        = $eventArgs->getEntity();
         $classMetadata = $em->getClassMetadata(get_class($entity));
 
-        if (!$this->getClassAnalyzer()->hasMethod($classMetadata->reflClass, 'setCurrentLocale')) {
+        if (!$this->isTranslatable($classMetadata)) {
             return;
         }
 


jordisala1991:
Ping @NiR-

michelcourtade:
Hi, 
same problem here with Sylius (1.2.10) managing his own translations and knp doctrine behaviors for our own entities in our project.
Any chance to see this PR merged soon ?
Thanks !

AlucardleVash:
Hi,
The next release will use interface instead checking trait usage. It's a BC Break, so it's should be an other major version number. It should fix your trouble too and release soon. We exchange between us to merge this one in v1. The fix work for most of user but some use their own implementation and this PR will break their usage.

michelcourtade:
Thanks for your response and the merge @AlucardleVash  !
Do you have a release date for next version ?

AlucardleVash:
Not yet but you can follow this require PR : 
https://github.com/KnpLabs/DoctrineBehaviors/pull/389
https://github.com/KnpLabs/DoctrineBehaviors/pull/387

michelcourtade:
Oh ok I'l take a look asap to see progress. By the way, good job on doctrine behavior 👍 !
Thanks

PR #407, opened 2018-12-28
[Translatable] avoids empty translations to be persisted

sangarbe:
When an empty translation is persisted, trying to update that translation afterwards with some content blows up a unique constraint violation. This is because current code doesn't return old empty translation (because it checks if it's empty inside "doTranslate" method), making new content to be interpreted like a brand new translation.

Currently, new empty translations are discarded properly, but updating a previous persisted translations can lead to a persisted empty translation, which produces the same effect as persisting brand new empty translations.

So, this change avoids any empty translation to be persisted in any case, fixing the constraint issue.


TomasVotruba:
Test case merged in #461

PR #413, opened 2019-02-04
[Translatable] make recursive mode configurable

vovkin:
When Translatable used in combination with inheritance map (in my specific case I'm using Single Table Inheritance), it does not work as expected for mapped entities (I got `null` istead of translations, because `TranslatableSubscriber::postLoad()` never called, and thus correct locale is not set).
 
I found the reason is that `TranslatableSubscriber::isTranslatable()` check returns false for mapped entities. It can be easily fixed by passing `$isRecursive` parameter as `true` to `ClassAnalyzer::hasTrait()`.

Proposed change in this PR will make `$isRecursive` of `TranlatableSubscriber` configurable.

I checked it in my project, and when proposed changes applied and in app config I set
```
parameters:
    knp.doctrine_behaviors.translatable_subscriber.is_recursive: true
```

translations works for mapped entities as expected.

### DIFF
--- a/config/orm-services.yml
+++ b/config/orm-services.yml
@@ -8,6 +8,7 @@ parameters:
     knp.doctrine_behaviors.translatable_subscriber.translation_trait: Knp\DoctrineBehaviors\Model\Translatable\Translation
     knp.doctrine_behaviors.translatable_subscriber.translatable_fetch_method: LAZY
     knp.doctrine_behaviors.translatable_subscriber.translation_fetch_method: LAZY
+    knp.doctrine_behaviors.translatable_subscriber.is_recursive: false
     knp.doctrine_behaviors.softdeletable_subscriber.class: Knp\DoctrineBehaviors\ORM\SoftDeletable\SoftDeletableSubscriber
     knp.doctrine_behaviors.softdeletable_subscriber.softdeletable_trait: Knp\DoctrineBehaviors\Model\SoftDeletable\SoftDeletable
     knp.doctrine_behaviors.timestampable_subscriber.class: Knp\DoctrineBehaviors\ORM\Timestampable\TimestampableSubscriber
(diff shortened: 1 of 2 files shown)


jerome-fix:
Hi,
I have the same issue with inherited entities.

It's a little strange to use traits for this. Maybe can we use an interface instead, and keeping the actual behavior (hasTrait) for not having BC.

I propose to change the isTranslatable method for : 

```
private function isTranslatable(ClassMetadata $classMetadata)
    {
        return  $this->getClassAnalyzer()->hasInterface($classMetadata->reflClass, TranslatableInterface::class)
            || $this->getClassAnalyzer()->hasTrait($classMetadata->reflClass, $this->translatableTrait);
    }
```

with : 

```
public function hasInterface(\ReflectionClass $class, string $interfaceName )
    {
        if (in_array($interfaceName, $class->getInterfaceNames())) {
            return true;
        }

        return false;
    }
```

Any translatable entity must implements `TranslatableInterface`. Need some documentation rewriting.

We could have the same comportment with method `isTranslation`

You can see my branch here : https://github.com/scc-digital/DoctrineBehaviors/tree/is-translatable

Regards,
Jérôme.

vovkin:
Hi @jerome-fix,

this should not be a BC, since `$isRecursive` is set as false by default, and that is its current value.

Also I think it's ok to rely on trait here, as that's how it works for other behaviors in this lib as well.

Proposed changes are minimal, and fit well in existing logic and style.

jerome-fix:
Hi @vovkin ,

You're right. Your proposal is more in the spirit of the current lib.
But I still find it awkward to rely on the use of traits. Maybe for a next major release?

Jérôme.

jaroslavlibal:
Is there any chance to get tis merged in the near future (major release), please? Or we'll have to stay on 1.5?

gsdevme:
We are seeing the same issue having updated to 1.6

PR #419, opened 2019-04-24
Fix for TreeBuilder E_USER_DEPRECATED message

deman714:
In /vendor/symfony/config/Definition/Builder/TreeBuilder.php:30
"A tree builder without a root node is deprecated since Symfony 4.2 and will not be supported anymore in 5.0."
Tested, works.


jkabat:
Why was this PR closed?

PR #421, opened 2019-05-22
Drop id property from translation trait

christingruber:
Dropped $id property from translation trait.

Any doctrine entity is shipped with its own $id property. Also, without the $id record inside the trait it is a little bit easier to deal with generated code like symfonys maker bundle or doctrine generator. 

Maybe for BC issues, an integration of mapId() is helpful:

https://github.com/KnpLabs/DoctrineBehaviors/blob/9bc5a6a6a6e6cb92ac7467410876882994732325/src/ORM/Translatable/TranslatableSubscriber.php#L93

### DIFF
--- a/src/Model/Translatable/TranslationMethods.php
+++ b/src/Model/Translatable/TranslationMethods.php
@@ -30,15 +30,6 @@ trait TranslationMethods
     }
 
     /**
-     * Returns object id.
-     *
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Thanks, I'll intergrate this feature in #442

PR #312, opened 2016-04-21
[Translatable] allow to add translatable behavior to abstract class and works with inheritance

almacbe:
There were some errors when you have some classes like the test that I have created and with the changes that I propuse I fix them.

What do you think?

PR #448, opened 2019-12-12
[Translation] add abstract class support

TomasVotruba:
Closes #312

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,8 +1,8 @@
 {
     "name": "knplabs/doctrine-behaviors",
-    "description": "Doctrine2 behavior traits",
+    "description": "Doctrine Behavior Traits",
     "keywords": [
-        "behaviors", "doctrine2", "timestampable", "translatable", "filterable", "blameable", "softdeletable", "tree"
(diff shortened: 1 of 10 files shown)


Kocal:
Hello :) 

To me, this PR fix obviously a bug, but it introduce a breaking change that has not been documented in the changelog.

In our project a work, which use KnpDoctrineBehaviors v1.5, we used to have:
- an abstract class `AbstractContent` (translatable)
- a class `AbstractContentTranslation` (translation)
- and many classes which extend `AbstractContent` (like `ArticleContent`, `PageContent`, ...).

It worked fine, `ArticleContent::getTranslationEntityClass()` returned `AbstractContentTranslation`, but since we upgraded to the last version it now returns `ArticleContentTranslation` which does not exist.

I will send a PR soon to document it in the changelog.

bobvandevijver:
Note that the new behavior also breaks the form when a Doctrine proxy object is fed as form data. I agree that shouldn't happen, but for anybody seeing the same behavior: this is the cause.

PR #452, opened 2019-12-13
Add missing dependency symplify/package-builder

### LINKED ISSUE #446: Feedback about 2.x@alpha
Hello,

Thanks for your recent hard work @TomasVotruba !

I updated quickly an old public repo of mine https://github.com/a2lix/Demo which used KnpDoctrineBehaviors 1.x and its Translatable feature to your latest 2.x@alpha, to test it and give you some feedback.

So, with the current alpha, there was a first error during the cache generation about:

```
!!  In DefinitionErrorExceptionPass.php line 54:
!!                                                                                 
!!    Cannot autowire service "Knp\DoctrineBehaviors\ORM\Geocodable\GeocodableSub  
!!    scriber": argument "$locationProvider" of method "__construct()" references  
!!     interface "Knp\DoctrineBehaviors\Contract\Provider\LocationProviderInterfa  
!!    ce" but no such service exists. Did you create a class that implements this  
!!     interface?        
```

That I solved for now, by exclude the entire directory Knp\DoctrineBehaviors\ORM\Geocodable of the autowiring.
(BTW about autowiring in public library/bundle, there was https://symfony.com/doc/current/service_container/autowiring.html#public-and-reusable-bundles but not sure if it is always up to date..)

See if you plan to handle back YAML configuration like https://github.com/a2lix/Demo/blob/master/app/config/packages/knplabs.yaml to only enable some features or if you have another idea in mind?

As a second point, currently, the `symplify/package-builder` dependancy is in use at https://github.com/KnpLabs/DoctrineBehaviors/blob/master/src/ORM/Translatable/TranslatableSubscriber.php#L194 and is missing in the composer "require" part. 
See if you want to keep it or have another plan.

And, it works :)
Thanks!

webda2l:
Close #446

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,7 @@
         "symfony/dependency-injection": "^4.4|^5.0",
         "symfony/security-core": "^4.4|^5.0",
         "symfony/cache": "^4.4|^5.0",
+        "symplify/package-builder": "^7.1",
         "nette/utils": "^3.0",
         "ramsey/uuid": "^3.9"
     },


TomasVotruba:
Thank you :+1:

PR #454, opened 2019-12-13
[WIP] Allo repository as a service

TomasVotruba:
Now only pure children of `EntityRepository` are supported, which is cruel vendor lock for such a simple features.

This PR should allow other repositories to use it:
https://www.tomasvotruba.cz/blog/2017/10/16/how-to-use-repository-with-doctrine-as-service-in-symfony/


TomasVotruba:
Replaced by https://github.com/KnpLabs/DoctrineBehaviors/pull/457

PR #468, opened 2019-12-19
Use symfony/strings instead of transliterator

TomasVotruba:


tests/ORM/SluggableTest.php
```diff
@@ -68,7 +68,7 @@ public function provideDataForTest(): Iterator
     {
         yield ['The name', 'the-name'];
         yield ['Löic & René', 'loic-rene'];
-        yield ['Иван Иванович', 'ivan-ivanovich'];
+        yield ['Иван Иванович', 'ivan-ivanovic'];
```
TomasVotruba: This was actually a bug. "ivanovic" is the right one

kdambekalns:
Checked `symfony/strings` as an alternative to `behat/transliterator` for [our project](https://github.com/neos/neos-development-collection/pull/2804) but ran into some differences, see https://github.com/symfony/symfony/issues/35061 - I thought this might be of interest to you.

TomasVotruba:
Thanks for sharing. I think that's expected from such a fresh library. Altought it's not perfect yet, it's stable and maintained, which is priority now.
I believe it becomes more stable in the future.

Issue #471, opened 2019-12-24
Sluggable generate unique slug bug


hermann8u:
Hi,

I have an issue with the v2.0.0-alpha4 with the generation of a unique Slug on the method Knp\DoctrineBehaviors\ORM\Sluggable\SluggableSubscriber::generateUniqueSlugFor.

This error occurs when I create my entity for the first time.

The issue is at line 84 because my slug is null at this moment and the interface define the getSlug method with on non nullable string as return type.

I defined my slug like this

 ```
    public function shouldGenerateUniqueSlugs(): bool
    {
        return true;
    }

    private function getRegenerateSlugOnUpdate(): bool
    {
        return false;
    }
```

## EDIT

If we look at the code of the *SluggableSubscriber*, we can see in the *processLifecycleEventArgs* that the method *generateSlug* is called. Because I put the result of getRegenerateSlugOnUpdate to false in my entity, the slug is never generated (see Knp\DoctrineBehaviors\Model\Sluggable\SluggableMethodsTrait:32 )

So finally, the bug is not related with uniqueness of the slug but with the getRegenerateSlugOnUpdate method when the result is set to false

TomasVotruba:
Hi, thanks for repoting this.

Any idea how to fix it?

hermann8u:
To me, the easiest way to fix it is to complete the first check I the generateSlug of the trait method. If the slug is null, we continue even if the 
getRegenerateSlugOnUpdate method return false.

I'll do a PR if you want in some days but for the moment it's Christmas time!

TomasVotruba:
> I'll do a PR if you want in some days but for the moment it's Christmas time!

:+1:

hermann8u:
This issue is fixed by the PR #472

PR #472, opened 2019-12-25
Fix slug generation if the getRegenerateSlugOnUpdate method return false

hermann8u:
This PR aims to fix the bug described in the issue #471 

If the getRegenerateSlugOnUpdate  method return false, the slug will be generate if its value is null. 

This behavior seems legit because you can set the slug manually and it won't be regenerated. However if you set manually the slug value to null, it will be regenerated (the setter doesn't allow this anyway).

### DIFF
--- a/tests/Fixtures/Entity/SluggableWithoutRegenerateEntity.php
+++ b/tests/Fixtures/Entity/SluggableWithoutRegenerateEntity.php
@@ -4,8 +4,6 @@ declare(strict_types=1);
 
 namespace Knp\DoctrineBehaviors\Tests\Fixtures\Entity;
 
-use DateTime;
-use DateTimeInterface;
 use Doctrine\ORM\Mapping as ORM;
 use Knp\DoctrineBehaviors\Contract\Entity\SluggableInterface;
 use Knp\DoctrineBehaviors\Model\Sluggable\SluggableTrait;
--- a/tests/Fixtures/Entity/SluggableWithoutRegenerateEntity.php
+++ b/tests/Fixtures/Entity/SluggableWithoutRegenerateEntity.php
@@ -31,17 +31,6 @@ class SluggableWithoutRegenerateEntity implements SluggableInterface
      */
     private $id;
 
-    /**
-     * @ORM\Column(type="datetime")
-     * @var DateTimeInterface
-     */
-    private $date;
-
-    public function __construct()
-    {
-        $this->date = (new DateTime())->modify('-1 year');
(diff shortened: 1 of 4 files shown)


TomasVotruba:
Wow, thank you for finding time! :christmas_tree: 

Could you also add test case to show, what went wrong? So we don't have this bug in the future due to unrelated changes.

hermann8u:
Any idea how to ignore this blocking "CS" check?

The getRegenerateSlugOnUpdate method is private on the trait but is considered as unused in the CS check when I override it in the implementing entity.

tests/Fixtures/Entity/SluggableWithoutRegenerateEntity.php
```diff
@@ -0,0 +1,82 @@
...
+     * @ORM\GeneratedValue(strategy="AUTO")
+     * @var int
+     */
+    private $id;
+
+    /**
+     * @ORM\Column(type="datetime")
+     * @var DateTimeInterface
+     */
+    private $date;
```
TomasVotruba: Where is this property used?

TomasVotruba:
> Ok, I just add the class to the skip section for this rule of the CS config

:+1: 

Thanks for adding test case here.

tests/Fixtures/Entity/SluggableWithoutRegenerateEntity.php
hermann8u: It is not but I keep it from the SluggableEntity I copied. The goal of this property is to have an other property which is not concerned by the slug generation. For the moment, this property is useless, so we have 3 choices:
- Keep it for future test case
- Remove it because there is no test based on it
- Add test concerning it like in Knp\DoctrineBehaviors\Tests\ORM\SluggableTest class

What do you think?

TomasVotruba: I'd remove it in both, because it doesn't add any value

hermann8u: The value of the createdAt property in Knp\DoctrineBehaviors\Tests\Fixtures\Entity\SluggableEntity is to test that the slug is not updated when we change an unrelated value.
But in Knp\DoctrineBehaviors\Tests\Fixtures\Entity\SluggableWithoutRegenerateEntity, there is no test with this createdAt property so I agree to delete it.

TomasVotruba:
Thank you :+1:

Issue #476, opened 2020-01-01
[Sluggable] SluggableRepository->isSlugUniqueFor() returns inversed boolean value


StanislavUngr:
**Proposal for fix**

https://github.com/KnpLabs/DoctrineBehaviors/blob/master/src/Repository/DefaultSluggableRepository.php#L22

Should return **true** if no result was found and **false** if any result with same slug was found. Now it is inversed and thus the logic of `generateUniqueSlugFor` in `SluggableSubscriber` doesnt work

I tried to make a PR but didnt get past new branch pushing..

TomasVotruba:
> I tried to make a PR but didnt get past new branch pushing..

What exactly do you mean?

To make PR in the most simple way, just edit the file here:

![ssssssss](https://user-images.githubusercontent.com/924196/71643636-5a60f980-2cbc-11ea-917a-8d8f02fe2f5a.png)

It's also possible you have outdated fork, from previous PR. In such case delete your forked repository first, then edit file here.

PR #477, opened 2020-01-01
Fix slug uniqueness check function

StanislavUngr:
Closes https://github.com/KnpLabs/DoctrineBehaviors/issues/476


TomasVotruba:
Thank you

PR #489, opened 2020-01-07
UuidablePropertiesTrait - uuid should be protected

lsv:
uuid - should be protected, so it can be overwritten.

### DIFF
--- a/src/Model/Uuidable/UuidablePropertiesTrait.php
+++ b/src/Model/Uuidable/UuidablePropertiesTrait.php
@@ -11,5 +11,5 @@ trait UuidablePropertiesTrait
     /**
      * @var UuidInterface|string|null
      */
-    private $uuid;
+    protected $uuid;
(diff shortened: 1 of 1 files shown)


TomasVotruba:
What's the real life use case?

lsv:
```
trait IdTrait {

    use UuidableTrait;

    /**
     * @ORM\Id
     * @ORM\Column(name="id", type="uuid", unique=true)
     * @ORM\GeneratedValue(strategy="CUSTOM")
     * @ORM\CustomIdGenerator(class="Ramsey\Uuid\Doctrine\UuidGenerator")
     */
    protected $uuid;
}
```

Why I use this trait, is because I dont want to type those 4 `ORM\...` on all my entities, and I can not change it if it is a `private` but works fine if its `protected`

TomasVotruba:
I see.

Thanks

PR #491, opened 2020-01-07
[Blameable] Added new parameter to allow definition of the user entity

### LINKED ISSUE #486: [Blameable] Issue with doctrine mapping
After installing v2.0 on an existing project and running the command `php bin/console doctrine:migrations:diff` I had many changes:
```sql
ALTER TABLE simple_pages DROP FOREIGN KEY FK_92660F3F896DBBDE;
ALTER TABLE simple_pages DROP FOREIGN KEY FK_92660F3FB03A8386;
ALTER TABLE simple_pages DROP FOREIGN KEY FK_92660F3FC76F1F52;
DROP INDEX IDX_92660F3FC76F1F52 ON simple_pages;
DROP INDEX IDX_92660F3FB03A8386 ON simple_pages;
DROP INDEX IDX_92660F3F896DBBDE ON simple_pages;
ALTER TABLE simple_pages ADD created_by VARCHAR(255) DEFAULT NULL, ADD updated_by VARCHAR(255) DEFAULT NULL, ADD deleted_by VARCHAR(255) DEFAULT NULL, DROP created_by_id, DROP updated_by_id, DROP deleted_by_id;
ALTER TABLE simple_pages_translations DROP FOREIGN KEY FK_B34E1BEA896DBBDE;
ALTER TABLE simple_pages_translations DROP FOREIGN KEY FK_B34E1BEAB03A8386;
ALTER TABLE simple_pages_translations DROP FOREIGN KEY FK_B34E1BEAC76F1F52;
DROP INDEX IDX_B34E1BEA896DBBDE ON simple_pages_translations;
DROP INDEX IDX_B34E1BEAB03A8386 ON simple_pages_translations;
DROP INDEX IDX_B34E1BEAC76F1F52 ON simple_pages_translations;
ALTER TABLE simple_pages_translations ADD updated_by VARCHAR(255) DEFAULT NULL, ADD deleted_by VARCHAR(255) DEFAULT NULL, DROP created_by_id, DROP updated_by_id, DROP deleted_by_id, CHANGE locale locale VARCHAR(5) NOT NULL, CHANGE slug created_by VARCHAR(255) DEFAULT NULL;
```

I should only have :
```sql
ALTER TABLE simple_pages_translations CHANGE locale locale VARCHAR(5) NOT NULL;
```

It's because this line https://github.com/KnpLabs/DoctrineBehaviors/blob/master/src/EventSubscriber/BlameableSubscriber.php#L142 returns `null` in cli context and `App\Entity\User` in http context.

CLI context: function `mapStringUser` (https://github.com/KnpLabs/DoctrineBehaviors/blob/master/src/EventSubscriber/BlameableSubscriber.php#L161) is used
HTTP context: function `mapManyToOneUser` (https://github.com/KnpLabs/DoctrineBehaviors/blob/master/src/EventSubscriber/BlameableSubscriber.php#L149) is used

This means even if I run the SQL queries above, I will have an issue because of this difference in the processing.

### My proposal

As previously done for blameable, there should be a parameter: `doctrine_behaviors_blameable_user_entity` equivalent of previous parameter `knp.doctrine_behaviors.blameable_subscriber.user_entity` in which the user class should be added.
Instead of using the UserProvider to update the class metadata info the parameter should be used.

I could be interesting to use an interface; so only classes implementing that interface would be allowed by the user provider.

fsoedjede:
Closes #486 

I can't find a way to get the user in the security provider. In order to make Blameable works well, a added the parameter `doctrine_behaviors_blameable_user_entity` which allows the user to define the entity to use

### DIFF
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -1,6 +1,7 @@
 parameters:
     doctrine_behaviors_translatable_fetch_mode: "LAZY"
     doctrine_behaviors_translation_fetch_mode: "LAZY"
+    doctrine_behaviors_blameable_user_entity: ~
 
 services:
     _defaults:
@@ -10,6 +11,7 @@ services:
         bind:
             $translatableFetchMode: "%doctrine_behaviors_translatable_fetch_mode%"
             $translationFetchMode: "%doctrine_behaviors_translation_fetch_mode%"
+            $blameableUserEntity: "%doctrine_behaviors_blameable_user_entity%"
 
     Knp\DoctrineBehaviors\:
         resource: "../src"
--- a/docs/blameable.md
+++ b/docs/blameable.md
@@ -48,3 +48,13 @@ $updatedBy = $category->getUpdatedBy();
 var_dump($updatedBy);
 // "App\Entity\User" object
 ```
+
+## Configuration
+
+By default, no user entity is provided. You need to specify the User class with a new parameter in your config:
+
+```yaml
+# services.yaml
+parameters:
+    doctrine_behaviors_blameable_user_entity: App\Entity\User
+```
--- a/src/EventSubscriber/BlameableSubscriber.php
+++ b/src/EventSubscriber/BlameableSubscriber.php
@@ -32,6 +32,11 @@ final class BlameableSubscriber implements EventSubscriber
     private const CREATED_BY = 'createdBy';
 
     /**
+     * @var string|null
+     */
+    private $blameableUserEntity;
+
+    /**
      * @var UserProviderInterface
      */
     private $userProvider;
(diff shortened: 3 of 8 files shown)


TomasVotruba:
Test is needed, to make sure this won't happen again.
Ideally failing CI first, than fix with CI passing.

fsoedjede:
Ok I will add it as soon as I have some time

spackmat:
I tested this PR and it solves the problem with the migrations well for my app.

For the record: Another symptom of the underlying problem is also solved by this PR: The error message `Error: Property "createdBy" in "App\Entity\MyEntity was already declared, but it must be declared only once` vanishes where MyEntity itself has no Blameable, but it is one of many extending MyBaseEntity using single table inheritance which uses the BlameableTrait and implements the BlameableInterface.

TomasVotruba:
@fsoedjede Thank you for your work and adding a test case

TomasVotruba:
@spackmat Thanks for feedback, that's great to hear. I've merged it now

spackmat:
@TomasVotruba I already had written a new issue for that problem (driving me nuts the whole day) but then realized that it is maybe related to the migrations/cli problem I also had and volià it is. One less issue. :) I don't have a clear concept of why this error vanishes, but I am quite happy it does so. So thanks to you both!

(Btw: v2.0.0 worked not at all for me creating segmentation faults on cache clear, memory exhaustion based 500s at the webinterface and freezing my whole VM on `doctrine:migrations:diff`)

TomasVotruba:
The segfaults was probably caused by https://github.com/KnpLabs/DoctrineBehaviors/pull/493

Issue #492, opened 2020-01-07
[Blameable] Maximum function nesting level of '256' reached


toooni:
I don't exactly know why but `BlameableSubscriber` creates an error "Maximum function nesting level of '256' reached" (segfault when xdebug is disabled). It seems like the issue can be solved for me by not using `$entityManager->getUnitOfWork()` directly in the constructor and calling `$this->entityManager->getUnitOfWork()` in the methods inside the subscriber.
I will be creating a PR as soon as https://github.com/KnpLabs/DoctrineBehaviors/pull/491 is merged.

TomasVotruba:
It's infinite loop.
Could you share minimal entity code to replicate?

TomasVotruba:
The referenced PR will need more time/work, so I would not wait for it

toooni:
There's no code to replicate. The issue happens for me without having a single `BlameableInterface`. I only use the `Translation`/`Translatable` stuff and successfully migrated from v1 to v2.
I think it will be better anyway if the `getUnitOfWork` call doesn't happen in the constructor. If you agree I will make a PR tomorrow morning.

TomasVotruba:
> I think it will be better anyway if the getUnitOfWork call doesn't happen in the constructor. If you agree I will make a PR tomorrow morning.

I see, yes :+1:

toooni:
Since you were so fast I was able to create the PR right now ;)

PR #493, opened 2020-01-07
Moved unitOfWork access to methods instead of constructor

toooni:
Fixes https://github.com/KnpLabs/DoctrineBehaviors/issues/492

For me this fixes a loop which made PHP crash. I am not sure why this happened but moving the unitOfWork access to the functions would make sense anyway.

### DIFF
--- a/src/EventSubscriber/BlameableSubscriber.php
+++ b/src/EventSubscriber/BlameableSubscriber.php
@@ -10,7 +10,6 @@ use Doctrine\ORM\Event\LifecycleEventArgs;
 use Doctrine\ORM\Event\LoadClassMetadataEventArgs;
 use Doctrine\ORM\Events;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
-use Doctrine\ORM\UnitOfWork;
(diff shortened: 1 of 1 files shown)


src/EventSubscriber/BlameableSubscriber.php
```diff
@@ -108,7 +107,7 @@ public function preUpdate(LifecycleEventArgs $lifecycleEventArgs): void
         $oldValue = $entity->getUpdatedBy();
         $entity->setUpdatedBy($user);
 
-        $this->unitOfWork->propertyChanged($entity, self::UPDATED_BY, $oldValue, $user);
+        $this->entityManager->getUnitOfWork()->propertyChanged($entity, self::UPDATED_BY, $oldValue, $user);
```
TomasVotruba: Could you extract `private function getUnitOfWork()` function, so we don't chained methods here?

toooni: done 👍

TomasVotruba:
Thank you :+1:

toooni:
Thanks to you! I just read your blog post. Awesome work 👍

PR #495, opened 2020-01-08
allow translatable to be cached in 2nd level cache

calvera:


TomasVotruba:
We need to cover it with minimal test case. In case the Doctrine changes

itarcontact:
Any updates here?

TomasVotruba:
Closing as stale

PR #498, opened 2020-01-11
Add another layer for locale fallback - %locale% parameter

toooni:
In the versions <2 the `%locale%` parameter was used as the fallback locale. Now we are trying to get the `defaultLocale` from the Request which is fine IMO. But we should use the `%locale%` parameter if no reqest exists.
I am aware that the `%locale%` parameter is now mandatory again. Is this a problem? Should we switch to `%kernel.default_locale%` which would require that we require `symfony/framework` as a dependency AFAIK.

Fixing this issue is mostly important because now there is no default locale set on cli.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,7 @@
         "symfony/dependency-injection": "^4.4|^5.0",
         "symfony/http-kernel": "^4.4|^5.0",
         "symfony/security-core": "^4.4|^5.0",
+        "symfony/framework-bundle": "^4.4|^5.0",
         "symfony/string": "^5.0",
         "symplify/package-builder": "^7.2",
         "nette/utils": "^3.0",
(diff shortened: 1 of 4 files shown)


TomasVotruba:
> Should we switch to %kernel.default_locale% which would require that we require symfony/framework as a dependency AFAIK.

I think `locale` parameter can be used, as it is the entry parameterfor symfony translation.

Maybe it would be better to avoid config passing some parameter and use Symfony service instead:
https://symfony.com/blog/new-in-symfony-4-1-getting-container-parameters-as-a-service

toooni:
@TomasVotruba Great idea. Didn't know about that. Now we have also a solution if the parameter is not set 👍

TomasVotruba:
This should help fix the test:

```diff
+"symfony/framework-bundle": "^4.4|^^5.0",
```

toooni:
👍 done

TomasVotruba:
Thank you for fast and clear work :+1:

PR #499, opened 2020-01-11
Address deprecations from persistence lib

greg0ire:
A backwards-compatibility layer has been added to persistence to help
consumers move to the new namespacing. It is based on class aliases,
which means the type declaration changes should not be a BC-break: types
are the same.
See doctrine/persistence#71

This means:
- using the new namespaces
- adding autoload calls for new types to types that may be extended and
use persistence types in type declarations of non-constructor methods,
so that signature compatibility is recognized by old versions of php.
Luckily, none of those were found in this package.
More details on this at
https://dev.to/greg0ire/how-to-deprecate-a-type-in-php-48cf

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
     "require": {
         "php": "^7.2",
         "doctrine/common": "^2.7",
-        "doctrine/persistence": "^1.3.3",
+        "doctrine/persistence": "^1.3.4",
         "doctrine/dbal": "^2.9",
         "doctrine/orm": "^2.7",
         "symfony/cache": "^4.4|^5.0",
--- a/tests/ORM/BlameableTest.php
+++ b/tests/ORM/BlameableTest.php
@@ -4,8 +4,8 @@ declare(strict_types=1);
 
 namespace Knp\DoctrineBehaviors\Tests\ORM;
 
-use Doctrine\Common\Persistence\ObjectRepository;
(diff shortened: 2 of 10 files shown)


TomasVotruba:
Thanks for the PR.

When this change happened first? It should be covered in Rector to instant upgrade of doctrine/persistance

greg0ire:
Here is a link to the PR: https://github.com/doctrine/persistence/pull/71

TomasVotruba:
Oh, sorry, I though it fixing some BC break, not moving.

So basically all `Doctrine\Common\*` classes from that package are now `Doctrine\Persistence\*` from version 1.4?

greg0ire:
Yes, _basically_

But take care, if you have any extensible type in your consuming library, you cannot really do the move without adding an autoload call so php does not crash when checking compatibility of extending types. Here is an example of such a call: https://github.com/doctrine/mongodb-odm/pull/2150/files#diff-7b9e4e14f6fec98822b6c1fb390ecbc4R480

Note that the call is useless if you are using php 7.4

TomasVotruba:
I rather think of version 2.0 migration path where the BC layer is removed.

greg0ire:
Oh right if it's compatibility with 2.0, it's as complicated as I said, but if it's migration to 2.0 (and you drop 1), it's simple.

PR #505, opened 2020-01-15
Merge pull request #505 from fsoedjede/488-translatable-with-sluggable

### LINKED ISSUE #488: [Sluggable + Translatable] Unique slug generatiion
Hi,

When using Sluggable with Translatable, there might be an issue if we want the slug to be unique.
Example:
`src/Entity/SimplePage.php`
```php
<?php
namespace App\Entity;
use Doctrine\ORM\Mapping as ORM;
use Knp\DoctrineBehaviors as ORMBehaviors;

class SimplePage implements ORMBehaviors\Contract\Entity\TranslatableInterface
{
    use ORMBehaviors\Model\Translatable\TranslatableTrait;
    /**
     * @var int
     * @ORM\Column(type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    private $id;

    /**
     * @return int
     */
    public function getId(): ?int
    {
        return $this->id;
    }
}
```

`src/Entity/SimplePageTranslation.php`
```php
<?php
namespace App\Entity;
use Doctrine\ORM\Mapping as ORM;
use Knp\DoctrineBehaviors as ORMBehaviors;

class SimplePageTranslation implements
    ORMBehaviors\Contract\Entity\SluggableInterface,
    ORMBehaviors\Contract\Entity\TranslationInterface
{
    use ORMBehaviors\Model\Sluggable\SluggableTrait;
    use ORMBehaviors\Model\Translatable\TranslationTrait;

    /**
     * @var int
     * @ORM\Id
     * @ORM\Column(type="integer")
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    private $id;
    /**
     * @var string
     * @ORM\Column(length=255)
     */
    private $title;

//Getter and setter
...
    public function getSluggableFields(): array
    {
        return ['title'];
    }
    public function shouldGenerateUniqueSlugs(): bool
    {
        return true;
    }
}
```

I'm using a2lix/translation-form-bundle so I have something like this:

image

I can save both french and english translations at the same time but if for exemple I enter for the title field `test` in both french and english, it will be save:
- fr => test
- en => test
instead of 
- fr => test
- en => test-1
because the insert is in the same transaction and the check for the uniqueness is done early: 
image

Standalone Sluggable does not have any problem
It's specific to Sluggable + Translatable usage

I can't currently see how this could be fixed. Does anyone have an idea?

fsoedjede:
Closes #488 

I use the UnitOfWork because that's the only side I could get entities which are going to be save and check if the slug does not match the current one

### DIFF
--- a/src/EventSubscriber/SluggableSubscriber.php
+++ b/src/EventSubscriber/SluggableSubscriber.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Knp\DoctrineBehaviors\EventSubscriber;
 
 use Doctrine\Common\EventSubscriber;
+use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Event\LifecycleEventArgs;
 use Doctrine\ORM\Event\LoadClassMetadataEventArgs;
 use Doctrine\ORM\Events;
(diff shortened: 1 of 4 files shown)


TomasVotruba:
Thank you :+1:

Issue #496, opened 2020-01-10
[Timestampable] Existing null values from v1 throw exceptions


spackmat:
I migrated a project from v1.6 to v2.0.1 an get a lot of Exeptions from the Timestampable Behavior:

> Return value of App\Entity\MyEntity::getUpdatedAt() must implement interface DateTimeInterface, null returned

Indeed, `getUpdatedAt()` and the other getters have a signature that allows only `DateTimeInterface` as a return value, but my existing data has always `null` for all entities that were never updated.

I'm not sure about side effects and why the signature was changed that way, but I think `?DateTimeInterface` as a return type hint would solve that.

TomasVotruba:
>  but I think ?DateTimeInterface as a return type hint would solve that.

:+1:

spackmat:
When I see such strict signatures, I always assume that it is was introduced on purpose, so in that case it maybe would be better to just return the value of `createdAt`, when `updatedAt` is `null`?

TomasVotruba:
It was on purpose within scope of this library's tsets, but it didn't cover all the possible cases. 

> Better to just return the value of createdAt, when updatedAt is null?

That could make sense.

spackmat:
That would be a BC break as I used updatedAt is null to mark entities as never updated for the users. I could change that, but it's not clear to me, why updatedAt should not be null. You say because the tests say so. But why did you change that at all from v1?

I made the return type hint nullable in the interface and the implementation and that works fine as far as I could see. That solution wouldn't be a BC break. Should I send a PR for that?

TomasVotruba:
Yes, I agreed to PR in previous comment:
https://github.com/KnpLabs/DoctrineBehaviors/issues/496#issuecomment-572951587

spackmat:
Here it is. But to be honest, I'm not sure, if the decision to always populate `updatedAt` is a good one.

The upside ist, that one only has to to look into that property and always get valid data.

The downsides are that it must have an inconsistent method signature to not break legacy data, stores redundant data for not-updated entities and it is a bit more complicated to check, if an entity was ever updated (besides that populating `updatedAt` is simply not correct, when an entity was in fact never updated).

So as an alternative to my PR I see two possible solutions: Update the legacy data to not have null values in the `updatedAt` column or reverting the decision and behave like v1 again, not touching `updatedAt` on persist.

TomasVotruba:
> ...the decision to always populate updatedAt is a good one.

[...]

PR #508, opened 2020-01-16
Allow updatedAt to be null to support legacy data

spackmat:
Fixing the BC break described in #496:  v1 had left `updatedAt` as `null`, if an entity was never updated. v2 introduces that `updatedAt` is initialized with the same timestamp as `createdAt`, so `updatedAt` cannot be `null` anymore. This PR changes the signature of `getUpdatedAt()` and `setUpdatedAt()` to allow null values again for legacy data (just to be consistent, as setting `updatedAt` to null will be overwritten on save).

### DIFF
--- a/src/Contract/Entity/TimestampableInterface.php
+++ b/src/Contract/Entity/TimestampableInterface.php
@@ -14,7 +14,7 @@ interface TimestampableInterface
 
     public function setCreatedAt(DateTimeInterface $createdAt): void;
 
-    public function getUpdatedAt(): DateTimeInterface;
+    public function getUpdatedAt(): ?DateTimeInterface;
 
-    public function setUpdatedAt(DateTimeInterface $updatedAt): void;
+    public function setUpdatedAt(?DateTimeInterface $updatedAt): void;
 }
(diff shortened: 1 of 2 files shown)


spackmat:
I didn't write a test for that because in the v2-world I cannot create a TimestampableEntity with `updatedAt` of `null`.

src/Contract/Entity/TimestampableInterface.php
```diff
@@ -14,7 +14,7 @@ public function getCreatedAt(): DateTimeInterface;
 
     public function setCreatedAt(DateTimeInterface $createdAt): void;
 
-    public function getUpdatedAt(): DateTimeInterface;
+    public function getUpdatedAt(): ?DateTimeInterface;
 
-    public function setUpdatedAt(DateTimeInterface $updatedAt): void;
+    public function setUpdatedAt(?DateTimeInterface $updatedAt): void;
```
TomasVotruba: Why this change?

spackmat: Just for consistency of the API. Doesn't make sense as I wrote in the description. I can remove that again, but the whole PR is just a suboptimal workaround, see my new comment in the issue #496

TomasVotruba:
I'm not sure what the problem is now. 
We need test to cover it.

In case of inconsistent date, I recommend using migration to complete them to database. We cannot compensate for broken data strutures here.

spackmat:
Forget that PR at all. It is a workaround for a BC break that either has to fixed with a migration of existing data from v1 or by reverting the BC break and restoring the old behavior.

PR #509, opened 2020-01-17
Use doctrine registry to cope with multiple managers

mduplouy:
Hi there,

This fix a MappingException (class not found in chain configured namespaces) when using several entity managers

### DIFF
--- a/src/Repository/DefaultSluggableRepository.php
+++ b/src/Repository/DefaultSluggableRepository.php
@@ -4,26 +4,30 @@ declare(strict_types=1);
 
 namespace Knp\DoctrineBehaviors\Repository;
 
+use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\ORM\EntityManagerInterface;
 use Knp\DoctrineBehaviors\Contract\Entity\SluggableInterface;
 
 final class DefaultSluggableRepository
 {
     /**
-     * @var EntityManagerInterface
+     * @var ManagerRegistry
      */
-    private $entityManager;
+    private $managerRegistry;
 
-    public function __construct(EntityManagerInterface $entityManager)
+    public function __construct(ManagerRegistry $managerRegistry)
     {
-        $this->entityManager = $entityManager;
+        $this->managerRegistry = $managerRegistry;
     }
 
     public function isSlugUniqueFor(SluggableInterface $sluggable, string $uniqueSlug): bool
     {
         $entityClass = get_class($sluggable);
 
-        $queryBuilder = $this->entityManager->createQueryBuilder()
+        /** @var EntityManagerInterface $entityManager */
+        $entityManager = $this->managerRegistry->getManagerForClass($entityClass);
+
+        $queryBuilder = $entityManager->createQueryBuilder()
             ->select('e')
             ->from($entityClass, 'e')
             ->select('COUNT(e)')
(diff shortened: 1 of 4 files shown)


src/Repository/DefaultSluggableRepository.php
```diff
@@ -4,26 +4,28 @@
...
+        $this->registry = $registry;
     }
 
     public function isSlugUniqueFor(SluggableInterface $sluggable, string $uniqueSlug): bool
     {
         $entityClass = get_class($sluggable);
 
-        $queryBuilder = $this->entityManager->createQueryBuilder()
+        $queryBuilder = $this->registry
+            ->getManagerForClass($entityClass)
```
TomasVotruba: Better decopule this logic to standalone line, as it is easy to oversee.

```php
$sluggableRepository = $this->registry->getManagerForClass($entityClass);
```

Also, this change should be done probably or all subscribers, right?

mduplouy: Just trying to get all test jobs ok before making any other modification :)

any idea on how to solve this one ? https://travis-ci.org/KnpLabs/DoctrineBehaviors/jobs/638486665

Do i really need to add a condition on the manager instance ?

TomasVotruba: @var annotation above the variable should solve that

mduplouy:
I finally made it :)

@TomasVotruba i also modified other subscribers, let me know if anything has to be modified

src/Repository/DefaultSluggableRepository.php
```diff
@@ -5,25 +5,29 @@
...
+        $this->managerRegistry = $managerRegistry;
     }
 
     public function isSlugUniqueFor(SluggableInterface $sluggable, string $uniqueSlug): bool
     {
         $entityClass = get_class($sluggable);
 
-        $queryBuilder = $this->entityManager->createQueryBuilder()
+        /** @var EntityManagerInterface $manager */
+        $manager = $this->managerRegistry->getManagerForClass($entityClass);
```
TomasVotruba:
```suggestion
        $entityManager = $this->managerRegistry->getManagerForClass($entityClass);
```

This would be more descriptive name :+1:

config/services.yaml
```diff
@@ -18,3 +18,5 @@ services:
         exclude:
             - "../src/Bundle/*"
             - "../src/Exception/*"
+
+    Doctrine\Persistence\ManagerRegistry: '@doctrine'
```
TomasVotruba: Why is this needed? It might break the app that is using this bundle

mduplouy: without this the autowire does not work well with some versions :/ 

https://travis-ci.org/KnpLabs/DoctrineBehaviors/jobs/638475665?utm_medium=notification&utm_source=github_status

TomasVotruba: I think it should be compatible with both. 
Moving alias to a services that's there in some conditions might break stuff.

TomasVotruba: Standalone test case for ManagerRegistry with own would definitelly help

mduplouy: Sorry but I don't understand what you mean
 imo autowiring should not be used and doctrine service explicitely injected

mduplouy: ok
apologize i though doctrine bundle wasn't a dev dependency
i found the weird alias stuff -- with if (false) condition ?? -- and type hinted on it, everything seems ok

mduplouy:
@TomasVotruba is this ok for you ?

TomasVotruba:
Code looks good. 
It needs test case with 2 entity managers to prove it works correctly

TomasVotruba:
Closing as stale

Issue #481, opened 2020-01-02
Trouble with make:entity


laurentHCM:
Hi Tomas,

I think there is just a small side-effect with the new release.
Each time I want to create an entity with bin/console make:entity , I have to following error message.
But if the entity is created, even empty, if I use again bin/console make:entity to provide fields to the entity , there is no more error message.
So the workaround is easy to manage.
Thanks.

php bin/console make:entity
Class name of the entity to create or update (e.g. BraveKangaroo):
 > BraveKangaroo

 created: src/Entity/BraveKangaroo.php
 created: src/Repository/BraveKangarooRepository.php

In BlameableSubscriber.php line 57:

  Call to a member function getName() on null

make:entity [-a|--api-resource] [--regenerate] [--overwrite] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-e|--env ENV] [--no-debug] [--] <command> [<name>]

TomasVotruba:
Hi,

what change in the subscriber you think would fix that?

laurentHCM:
Hi Tomas,

I will try to find and be back soon.

De : Tomas Votruba [mailto:notifications@github.com]
Envoyé : jeudi 2 janvier 2020 17:17
À : KnpLabs/DoctrineBehaviors
Cc : laurent BARBEAU; Author
Objet : Re: [KnpLabs/DoctrineBehaviors] Trouble with make:entity (#481)


Hi,

what change in the subscriber you think would fix that?

laurentHCM:
Hi Tomas,

Maybe in all subscribers we have to had a test to check if reflClass is not null , like below for BlameableSubscriber .
if (null == $classMetadata->reflClass || ! is_a($classMetadata->reflClass->getName(), BlameableInterface::class, true)) {

Do you think it’s enough ?

De : Tomas Votruba [mailto:notifications@github.com]
Envoyé : jeudi 2 janvier 2020 17:17
À : KnpLabs/DoctrineBehaviors
Cc : laurent BARBEAU; Author
Objet : Re: [KnpLabs/DoctrineBehaviors] Trouble with make:entity (#481)


Hi,

what change in the subscriber you think would fix that?

TomasVotruba:
Hi, 

that might be a fix. I still wonder why it's null and not the class name.
Could you think of any difference to your other entities?

laurentHCM:
[...]

PR #510, opened 2020-01-19
Make:entity dispatches event with incomplete argument

ewbarnard:
Resolves issue #481 by checking for a null reflection class pointer before
using the pointer, in each of the event subscribers.


TomasVotruba:
Thank you

PR #514, opened 2020-01-27
changelog: note breaking change about abstract translatable

Kocal:
See https://github.com/KnpLabs/DoctrineBehaviors/pull/448#issuecomment-578937662, thanks. :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,21 @@ PRs and issues are linked, so you can find more about it. Thanks to [ChangelogLi
 - [#463] drop filterable, way to opinionated and limited, use custom implementation
 - [#457] remove repository traits, use custom methods in own repository instead
 
+### BREAKING CHANGES
+
+#### [#448] [Translation] add abstract class support
+
+Given three classes:
+- `AbstractEntity` (translatable)
+- `AbstractEntityTranslation` (translation)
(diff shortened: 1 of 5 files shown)


TomasVotruba:
This could be handled better by instant upgrade with Rector.

What is the minimal diff that shows the change 1 to 2?

Kocal:
Hum that's true, but I don't really think this can by handled properly by rector.

This is how we _fixed_ it, we keep the old `AbstractContentTranslation` class:

```diff
<?php

// ...

abstract class AbstractContent implements TranslatableInterface
{
    use TranslatableTrait;

+   public static function getTranslationEntityClass(): string
+   {
+       return self::class.'Translation';
+   }

    // ...
}
```

TomasVotruba:
I see. Should be done.

 Could you add failing test case to repository here? I'll fix it and the prepare migration path

Kocal:
I'm not sure of what I should really do here. :/
I've added some fixtures (abstract entity and its translation in _v1 way_) and a test, but I'm facing `General error: 1 no such table: AbstractV1TranslatableEntityTranslation` error :(

TomasVotruba:
I'm not sure either, I only maintain this package for a few months and I'm not very active translation user.

TomasVotruba:
Closing as stale

Issue #516, opened 2020-02-03
Adding generics support 


Kocal:
Hi, :hand:  

This is a feature request for adding generics types for PHPStan (or Psalm) on `Translatable`/`Translation` interfaces:
- https://medium.com/@ondrejmirtes/generics-in-php-using-phpdocs-14e7301953
- https://arnaud.le-blanc.net/post/phpstan-generics.html

Given the following classes:
```php
<?php

namespace App\Entity;

use \Knp\DoctrineBehaviors\Contract\Entity\TranslatableInterface;
use \Knp\DoctrineBehaviors\Contract\Entity\TranslationInterface;
use \Knp\DoctrineBehaviors\Model\Translatable\TranslatableTrait;
use \Knp\DoctrineBehaviors\Model\Translatable\TranslationTrait;

/**
 * @implements TranslatableInterface<ArticleTranslation>
 */
class Article implements TranslatableInterface {
    use TranslatableTrait;
}

/**
 * @implements TranslationInterface<Article>
 */
class ArticleTranslation implements TranslationInterface {
    use TranslationTrait;

    public function getTitle() { /* ... */ }
    public function setTitle(string $title) { /* ... */ }
}

class FooBar {}
```

By using generic types, we will improve static analyze and developer experience (when IDEs will support PHPStan/Psalm generics syntax):

```php
<?php

$article = new Article();

$article->getTranslations(); // typed to a Collection of `ArticleTranslation`
$article->getNewTranslations(); // typed to a Collection of `ArticleTranslation`

$article->translate('fr'); // typed to `ArticleTranslation`
$article->translate('fr')->setTitle('Mon titre'); // ok
$article->translate('fr')->setFooBar('Mon titre'); // fail

$article->addTranslation(new ArticleTranslation); // ok
$article->addTranslation(new FooBar); // fail

$article->removeTranslation(new ArticleTranslation); // ok
$article->removeTranslation(new FooBar); // fail

$article->setTranslations(new ArrayCollection([new ArticleTranslation]); // ok
$article->setTranslations(new ArrayCollection([new ArticleTranslation, new FooBar]); // fail

$articleTranslation = /* ... */;
$articleTranslation->getTranslatable(); // typed to `Article`
```

---

On one of our project at work, I've started to implement generics myself (on `translate` method, with new traits and interfaces which extend code from vendor) but it didn't work properly. 

[...]

PR #517, opened 2020-02-04
PHPStan: generics on Transatable/Translation

Kocal:
This PR is a proposal for #516. 

The following code:
Capture d’écran de 2020-02-04 07-09-01

Will now be analysed correctly by PHPStan:
Capture d’écran de 2020-02-04 07-09-19

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -44,8 +44,5 @@ parameters:
 
         - '#Cannot call method addChildNode\(\) on Knp\\DoctrineBehaviors\\Contract\\Entity\\TreeNodeInterface\|null#'
 
-        # iterable
-        - '#Parameter \#1 \$translations of method Knp\\DoctrineBehaviors\\Tests\\Fixtures\\Entity\\TranslatableEntity\:\:setTranslations\(\) expects Doctrine\\Common\\Collections\\Collection&iterable<Knp\\DoctrineBehaviors\\Contract\\Entity\\TranslationInterface\>, array<int, Knp\\DoctrineBehaviors\\Contract\\Entity\\TranslationInterface\> given#'
-
         - '#Property Knp\\DoctrineBehaviors\\Provider\\LocaleProvider\:\:\$translator has no typehint specified#'
         - '#PHPDoc tag @var has invalid value \(TranslatorInterface&LocaleAwareInterface\|null\)\: Unexpected token "\|", expected TOKEN_OTHER at offset 56#'
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -11,9 +11,6 @@ parameters:
     reportUnmatchedIgnoredErrors: false
 
     ignoreErrors:
-        # traits
-        - '#Call to an undefined method Knp\\DoctrineBehaviors(.*?)#'
-
         # buggy
         - '#of function call_user_func_array expects callable#'
 
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -11,6 +11,21 @@ parameters:
     reportUnmatchedIgnoredErrors: false
 
     ignoreErrors:
+        # traits
+        - '#Call to an undefined method Knp\\DoctrineBehaviors\\Contract\\Provider\\UserProviderInterface::changeUser\(\)#'
+        - '#Call to an undefined method Knp\\DoctrineBehaviors\\Contract\\Entity\\SluggableInterface::getId\(\)#'
+        - '#Call to an undefined method Knp\\DoctrineBehaviors\\Contract\\Entity\\TreeNodeInterface::getId\(\)#'
+        - '#Call to an undefined method Knp\\DoctrineBehaviors\\Contract\\Entity\\TreeNodeInterface::to(Flat)?Array\(\)#'
+
+        # translation proxy
+        - message: '#Call to an undefined method Knp\\DoctrineBehaviors\\Tests\\Fixtures\\Entity\\TranslatableEntity::(get|set)Title\(\)#'
+          path: tests/ORM/TranslatableTest.php
+
+        # translation inheritance, I don't wanted to change (Abstract|Extended)TranslationEntity(Translation)? classes
+        # to had generics support because it feels super weird and inheritance was legit useless.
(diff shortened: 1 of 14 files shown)


src/Model/Translatable/TranslatableMethodsTrait.php
```diff
@@ -130,7 +131,7 @@ public static function getTranslationEntityClass(): string
      *
      * @param string $locale The locale (en, ru, fr) | null If null, will try with current locale
      */
-    protected function doTranslate(?string $locale = null, bool $fallbackToDefault = true): TranslationInterface
+    protected function doTranslate(?string $locale = null, bool $fallbackToDefault = true)
```
Kocal: Not a big fan, but it prevents the following issue:

Capture d’écran de 2020-02-04 07-18-40

Also this is a protected method which is only consumed by `translate()` (which still have `TranslationInterface` php return type)

src/Model/Translatable/TranslatableMethodsTrait.php
```diff
@@ -163,7 +164,6 @@ protected function doTranslate(?string $locale = null, bool $fallbackToDefault =
 
         $class = static::getTranslationEntityClass();
 
-        /** @var TranslationInterface $translation */
```
Kocal: Prevent this issue:
Capture d’écran de 2020-02-04 07-20-24


Not a big deal, this modification is inside `doTranslate()` method which is protected

src/Model/Translatable/TranslationPropertiesTrait.php
```diff
@@ -15,7 +15,7 @@ trait TranslationPropertiesTrait
 
     /**
      * Will be mapped to translatable entity by TranslatableSubscriber
-     * @var TranslatableInterface
+     * @var mixed
```
Kocal: Prevent this issue:

Capture d’écran de 2020-02-04 07-21-33

Not a big fan either, but this property is only used by `getTranslatable()` which has:
- php return type to `TranslatableInterface`
- phpstan return type to `TranslatableInterface<Translatable>`

phpstan.neon
```diff
@@ -12,7 +12,19 @@ parameters:
...
+        - '#Call to an undefined method Knp\\DoctrineBehaviors\\Contract\\Provider\\UserProviderInterface::changeUser\(\)#'
+        - '#Call to an undefined method Knp\\DoctrineBehaviors\\Contract\\Entity\\SluggableInterface::getId\(\)#'
+        - '#Call to an undefined method Knp\\DoctrineBehaviors\\Contract\\Entity\\TreeNodeInterface::getId\(\)#'
+        - '#Call to an undefined method Knp\\DoctrineBehaviors\\Contract\\Entity\\TreeNodeInterface::to(Flat)?Array\(\)#'
+
+        # translation proxy
+        - message: '#Call to an undefined method Knp\\DoctrineBehaviors\\Tests\\Fixtures\\Entity\\TranslatableEntity::(get|set)Title\(\)#'
+          path: tests/ORM/TranslatableTest.php
+
+        # translation inheritance, I don't wanted to change (Abstract|Extended)TranslationEntity(Translation)? classes
```
Kocal: Prevent this issue: 
Capture d’écran de 2020-02-04 07-25-14

Using `@implements` on abstract classes/children classes does not work very well. To make it work, some modifications are necessary but they are not pertinent and wanted I guess:
- `AbstractTranslatableEntity` should not implement `TranslatableInterface` and use trait `TranslatableTrait`
- `AbstractTranslatableEntityTranslation` should not implement `TranslationInterface` and use trait `TranslationTrait`
- `ExtendedTranslatableEntity` should implement `TranslatableInterface`, use trait `TranslatableTrait` and add `@implements` tag
- `ExtendedTranslatableEntityTranslation` should implement `TranslationInterface`, use trait `TranslationTrait` and add `@implements` tag

And so... inheritance now become useless :sweat_smile:

src/Model/Translatable/TranslatableMethodsTrait.php
```diff
@@ -11,9 +11,6 @@
 
 trait TranslatableMethodsTrait
 {
-    /**
-     * @return Collection|TranslationInterface[]
-     */
     public function getTranslations()
```
TomasVotruba: This should be returned, as it weakens PHPStorm autocomplete abilities

src/Model/Translatable/TranslatableMethodsTrait.php
TomasVotruba: Don't let tools force you to write bad code just to make them happy.
This destroys contract completely, must be reverted.

src/Model/Translatable/TranslatableMethodsTrait.php
TomasVotruba: Must be reverted

src/Model/Translatable/TranslationPropertiesTrait.php
TomasVotruba: Same thing, must be reverted to keep contract

src/Model/Translatable/TranslatableMethodsTrait.php
Kocal: We already have the PHPDoc on the [TranslatableInterface](https://github.com/Kocal/DoctrineBehaviors/blob/generics/translatable-translation/src/Contract/Entity/TranslatableInterface.php#L14-L18), it does not break PHPStorm autocomplete:
Capture d’écran de 2020-02-11 21-34-12

src/Model/Translatable/TranslatableMethodsTrait.php
Kocal: `doTranslate()` is a protected method which is not part of `TranslatableInterface` interface, so the contract is not broken. 
The method is only used by public method `translate()` which still have `TranslatableInterface` return type, so I think it's okay?

src/Model/Translatable/TranslatableMethodsTrait.php
TomasVotruba: How about human readability?

Kocal:
Hum, I'm really not proud of this PR. At first I _just_ wanted to add generics support with PHPStan but I had to modifiy too much PHP code...

I will close this PR and open another PR for a PHPStan extension. The final user will be able to use it like this:
```neon
includes:
    - vendor/knplabs/doctrine-behaviors/extension.neon
```

PR #523, opened 2020-02-29
[Timestampable] Move updateTimestamps method to UpdateTimestampsTrait

WolfgangDrescher:
Working with e.g. API Platform it may be useful to implement the createdAt and updatedAt methods of the TimestampableInterface yourself to add serialization group with annotations and not having to use YAML files instead for the configuration. Therefore it may be useful if KnpLabs DoctrineBehaviors would offer a trait for just the updateTimestamps method so we dont have to reimplement it ourselves in every project but can use the method of this package that already exists.

Maybe `UpdateTimestampsMethodTrait` will be better than `UpdateTimestampsTrait` as the trait name?

### DIFF
--- a/src/Model/Timestampable/TimestampableMethodsTrait.php
+++ b/src/Model/Timestampable/TimestampableMethodsTrait.php
@@ -4,13 +4,12 @@ declare(strict_types=1);
 
 namespace Knp\DoctrineBehaviors\Model\Timestampable;
 
-use DateTime;
 use DateTimeInterface;
-use DateTimeZone;
-use Knp\DoctrineBehaviors\Exception\ShouldNotHappenException;
 
 trait TimestampableMethodsTrait
 {
+    use UpdateTimestampsTrait;
+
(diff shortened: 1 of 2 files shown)


TomasVotruba:
This should be done at main trait level, to keep tree hierarchy

WolfgangDrescher:
```php
trait TimestampableTrait
{
    use TimestampablePropertiesTrait;
    use TimestampableMethodsTrait;
    use UpdateTimestampsTrait;
}
```

You mean like so and remove `use UpdateTimestampsTrait;` from `TimestampableMethodsTrait.php`? Not sure if this make sense tough since this could lead to unexpected behavior and break backward compatibility when just using `TimestampableMethodsTrait` and now the `updateTimestamps` method is missing.

TomasVotruba:
Split `TimestampableTrait` into update/created one

TomasVotruba:
What's the status here? Looks dead

TomasVotruba:
Closing as stale

PR #524, opened 2020-03-02
WIP: Mapping agnostic behaviors (adds Mongo ODM support)

alterphp:
By using Doctrine persistence interfaces, I think this bundle can handle ORM as well as Mongo ODM objects with a minimum of changes.

Here is a try for Timestampable behavior. Will you study this kind of PR ?

### DIFF
--- a/src/EventSubscriber/AbstractEventSubscriber.php
+++ b/src/EventSubscriber/AbstractEventSubscriber.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Knp\DoctrineBehaviors\EventSubscriber;
+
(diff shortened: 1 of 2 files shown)


TomasVotruba:
I don't think mixing 2 different usages of Doctrine (the original organization have it separated in ORM and ODM) will lead to healthy maintainable package.

I recommend you to create a new standalone package, that handles ODM. Eg. DoctrineODMBehaviors.

alterphp:
Ok, I understand your point of view. Sustainable maintenance is more important ;-)

TomasVotruba:
Thanks


Feel free to share the link to package here

PR #531, opened 2020-03-16
Fix bug in isIndirectChildNodeOf

martinknor:
materialized path /root/1 matched as indirect child node for /root/15, because strpos matched. Materialized path separator was add to fix this without bc break.


TomasVotruba:
To fix cs, just run:

```bash
composer fix-cs
```

martinknor:
I dont have install php 7.2, it would be good have docker container in repo for this :)

TomasVotruba:
Ah, you use the package in Docker?

> It would be good have docker container in repo for this :)

I don't use Docker. Feel free to send PR with one

martinknor:
Now cs in ok, but there is soNow cs in ok, but there is some other problem with php stan in tests... :(me other problem with php stan in tests... :(

TomasVotruba:
I'm going through the old issues. 
Closing this as test case is missing.

Thank you for your work.

PR #534, opened 2020-03-18
Sympony bandle name conventions

uginroot:
Package changes for the correct operation of symfony-flex auto-generated recipe

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,14 +54,14 @@
         }
     },
     "scripts": {
-        "check-cs": "vendor/bin/ecs check --ansi",
-        "fix-cs": "vendor/bin/ecs check --fix --ansi",
-        "phpstan": "vendor/bin/phpstan analyse --ansi --error-format symplify",
-        "rector": "vendor/bin/rector process --config rector-ci.yaml",
+        "check-cs": "\"vendor/bin/ecs\" check --ansi",
+        "fix-cs": "\"vendor/bin/ecs\" check --fix --ansi",
+        "phpstan": "\"vendor/bin/phpstan\" analyse --ansi --error-format symplify",
+        "rector": "\"vendor/bin/rector\" process --config rector-ci.yaml",
         "changelog": [
-            "vendor/bin/changelog-linker dump-merges --in-categories --ansi",
-            "vendor/bin/changelog-linker link --ansi",
-            "vendor/bin/changelog-linker cleanup --ansi"
+            "\"vendor/bin/changelog-linker\" dump-merges --in-categories --ansi",
+            "\"vendor/bin/changelog-linker\" link --ansi",
+            "\"vendor/bin/changelog-linker\" cleanup --ansi"
         ]
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -54,14 +54,14 @@
         }
     },
     "scripts": {
-        "check-cs": "\"vendor/bin/ecs\" check --ansi",
-        "fix-cs": "\"vendor/bin/ecs\" check --fix --ansi",
-        "phpstan": "\"vendor/bin/phpstan\" analyse --ansi --error-format symplify",
-        "rector": "\"vendor/bin/rector\" process --config rector-ci.yaml",
+        "check-cs": "vendor/bin/ecs check --ansi",
+        "fix-cs": "vendor/bin/ecs check --fix --ansi",
+        "phpstan": "vendor/bin/phpstan analyse --ansi --error-format symplify",
+        "rector": "vendor/bin/rector process --config rector-ci.yaml",
         "changelog": [
-            "\"vendor/bin/changelog-linker\" dump-merges --in-categories --ansi",
-            "\"vendor/bin/changelog-linker\" link --ansi",
-            "\"vendor/bin/changelog-linker\" cleanup --ansi"
+            "vendor/bin/changelog-linker dump-merges --in-categories --ansi",
+            "vendor/bin/changelog-linker link --ansi",
+            "vendor/bin/changelog-linker cleanup --ansi"
         ]
     }
 }
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -17,4 +17,5 @@ services:
         resource: "../src"
         exclude:
             - "../src/Bundle/*"
+            - "../src/DoctrineBehaviorsBundle.php"
             - "../src/Exception/*"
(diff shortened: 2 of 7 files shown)


tests/ORM/UuidableTest.php
```diff
@@ -26,7 +25,7 @@ public function testUuidLoading(): void
 
         $uuidableRepository = $this->entityManager->getRepository(UuidableEntity::class);
 
-        /** @var UuidableTrait $entity */
+        /** @var UuidableEntity $entity */
```
TomasVotruba:
```suggestion
        /** @var UuidableInterface $entity */
```

This would be correct, as Interface are source of truth. Trait are one of way to use them, but it should work without them.

TomasVotruba:
Change is needed, I miss-clicked wrong button.

composer.json
```diff
@@ -54,14 +54,14 @@
...
+        "fix-cs": "\"vendor/bin/ecs\" check --fix --ansi",
+        "phpstan": "\"vendor/bin/phpstan\" analyse --ansi --error-format symplify",
+        "rector": "\"vendor/bin/rector\" process --config rector-ci.yaml",
         "changelog": [
-            "vendor/bin/changelog-linker dump-merges --in-categories --ansi",
-            "vendor/bin/changelog-linker link --ansi",
-            "vendor/bin/changelog-linker cleanup --ansi"
+            "\"vendor/bin/changelog-linker\" dump-merges --in-categories --ansi",
+            "\"vendor/bin/changelog-linker\" link --ansi",
+            "\"vendor/bin/changelog-linker\" cleanup --ansi"
```
TomasVotruba: Why this? Should be reverted

TomasVotruba:
Could you share source for this convention with bundle class positioning? It's new to me

composer.json
uginroot: cmd

tests/ORM/UuidableTest.php
uginroot: Of course, you can remove this commit and add an exception for phpstan. But this option is also correct and does not require exceptions.

uginroot:
[Agreement](https://symfony.com/doc/current/bundles/best_practices.html#bundles-naming-conventions).
My [attempt ](https://github.com/symfony/recipes-contrib/pull/877)to add a recipe

composer.json
TomasVotruba: There is trick for this to shorten it to part after `bin`.
But that breaks while copying the command and modifying it. So this it the final choice.

TomasVotruba:
tl;dr; Where exactly is menti the path location?

uginroot:
https://symfony.com/doc/current/bundles/best_practices.html#bundle-name

TomasVotruba:
Still don't see it.

uginroot:
[Directory Structure](https://symfony.com/doc/current/bundles/best_practices.html#directory-structure). The documentation is incomplete, I had to delve into the symfony-flex code to understand where the file should be and what it should be called. symfony-flex search bundle file in psr-4 and psr-0 autoload directories. For this bundle simfony-flex check next choice:
```
class: DoctrineBehaviorsBundle 
path: /vendor/knplabs/doctrine-behaviors/src/DoctrineBehaviorsBundle.php

class: KnpDoctrineBehaviorsBundle
path: /vendor/knplabs/doctrine-behaviors/src/KnpDoctrineBehaviorsBundle.php
```
Patterns path:
```
[src]/[vendor][packege]Bundle.php
[src]/[packege]Bundle.php
```

TomasVotruba:
I can't see any information about required location in the root directory.

uginroot:
I also did not understand anything from the documentation and I had to look at the code. In place of symfony developers, I would use the recipe.json file directly in the bundle repository. So you would not have to create a separate repository with all recipes for all bundles and introduce restrictions related to the ban on adding a recipe with only one command adding a to the config/bandles.php file.

[All recipe](https://github.com/symfony/recipes-contrib)

About how the [recipe.json](https://github.com/symfony/recipes/blob/master/README.rst#configurators) file might look

TomasVotruba:
The bundle part is ok :+1: Thanks for information.

uginroot:
> Change is needed, I miss-clicked wrong button.

What needs to be done to get the pull request into the master branch?

TomasVotruba:
https://github.com/KnpLabs/DoctrineBehaviors/pull/534#discussion_r394883898

config/services.yaml
```diff
@@ -17,4 +17,6 @@ services:
         resource: "../src"
         exclude:
             - "../src/Bundle/*"
+            - "../src/DoctrineBehaviorsBundle.php"
             - "../src/Exception/*"
+            - "../src/PHPStan/*"
```
Kocal: There is an issue and a PR for that #537 #538, sorry :sweat_smile:

TomasVotruba:
Thank you :+1:

flug:
Be careful with this pr you with a bc break of namespace.

bobdenotter:
This change breaks all Bolt 4 builds. The next time you introduce a breaking change like this, please do a major version bump!

TomasVotruba:
Could send BC class?

bobdenotter:
Hi @TomasVotruba,

I'm not sure what you mean, but this was the change required on our side: https://github.com/bolt/core/pull/1249/files

Luckily there aren't many users of Bolt 4 in production yet, because it's still prerelease. But, if an unsuspected user runs `composer update` on their installation, they'll bump this library from 2.0.1 to 2.0.2, and thereby breaking their installation until they've fixed the `config/bundles.php`.

TomasVotruba:
@bobdenotter Some back compatibility class, that we could add and that covers both cases.

bobdenotter:
@TomasVotruba Maybe, but that would require it to be picked up by the autoloader. 

Alternatively, I tried doing this: 

```php
class_alias(\Knp\DoctrineBehaviors\DoctrineBehaviorsBundle::class, 'Knp\DoctrineBehaviors\Bundle\DoctrineBehaviorsBundle');
```

It does work, somewhat, but only if I put it in _our_ code, and it needs to be 'loaded' _before_ Symfony initializes the bundles. So, in `Kernel.php`, for example. It's not a pretty solution, though.. 
However, if _you_ would add a BC class now, that would clutter up the namespace for other users of this Library. 

Maybe wait a few days to see if others bump into this issue? If it's just us, we can deal with it in _our_ code for now :-)

TomasVotruba:
Better than class aliases magic (it's very hard to work with), is extending new class with deprecated one:

```php
/**
 * @deprecated
 */
final class OldOne extends NewOne
{
    // no code
}

class NewOne
{
    // ...
}
```

It's commonly use by Symfony

PR #538, opened 2020-03-24
fix(config): move PHPStan folder, fix #537

### LINKED ISSUE #537: PHPStan error on cache clear
Hi.

If I use dev-master in my repo, there is an issue because I don't use PHPStan.

```
Executing script cache:clear [KO]
 [KO]
Script cache:clear returned with error code 1
!!  
!!  In DefinitionErrorExceptionPass.php line 54:
!!                                                                                 
!!    Class "PHPStan\Type\DynamicMethodReturnTypeExtension" not found while loadi  
!!    ng "Knp\DoctrineBehaviors\PHPStan\Type\TranslatableGetTranslationsDynamicMe  
!!    thodReturnTypeExtension".                                                    
!!                                                                                 
!!  
!!  
Script @auto-scripts was called via post-update-cmd
```

Kocal:
Fix #537, ping @TomasVotruba 

PHPStan check failure is unrelated.


TomasVotruba:
CI needs fix

TomasVotruba:
Also, location of these classes is not correct. It's not part of `src` for this bundle.

Better location would be something like `utils/phpstan-behaviors/src`.

composer.json
```diff
@@ -50,7 +50,8 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "Knp\\DoctrineBehaviors\\Tests\\": "tests"
+            "Knp\\DoctrineBehaviors\\Tests\\": "tests",
+            "Knp\\DoctrineBehaviors\\PHPStan\\": "utils/phpstan-behaviors/src"
```
TomasVotruba: How can this work for 3rd party install?

Kocal:
I moved PHPStan code to `utils/phpstan-behaviors/src`, the namespace is still the same `Knp\DoctrineBehaviors\PHPStan`).

However for the CI issue, I can't reproduce on local and the code seems valid to me..

TomasVotruba:
CI still needs fixing/commenting out. I know it's unrelated, but CI should stay green

composer.json
Kocal: For a moment I thought `autoload-dev` would work in 3rd party :sweat_smile: 
Updated.

TomasVotruba:
Nice one :+1:

PR #539, opened 2020-03-26
Support ramsey/uuid version 4

tsctao:


TomasVotruba:
Better use 4.0 to keep consistent version

tsctao:
> Better use 4.0 to keep consistent version format

I followed [ramsey/uuid's documentation](https://uuid.ramsey.dev/en/latest/upgrading/3-to-4.html).

If we force version 4, it will conflict with user's projects that are not ready to upgrade ramsey/uuid to version 4?

composer.json
```diff
@@ -24,7 +24,7 @@
         "symfony/string": "^5.0",
         "symplify/package-builder": "^7.2",
         "nette/utils": "^3.0",
-        "ramsey/uuid": "^3.9"
+        "ramsey/uuid": "^3.9|^4"
```
Kocal:
```suggestion
        "ramsey/uuid": "^3.9|^4.0"
```

tsctao:
Oh, I misunderstood. Thank you.

TomasVotruba:
@tsctao I was on mobile so I didn't realize I could use suggestion. Code is always better than words.

Thank you both!

PR #541, opened 2020-03-30
[Sluggable] Enable gedmo-like "unique_base" feature

crazy-weasel:
Hi!
I'm migrating away from the Gedmo-DoctrineExtension and noticed that there is no equivalent to the unique_base option of Sluggable.
Switching the DefaultSluggableRepository would be fine for my usecase, but that's not possible because the class is marked as final.

I created an interface for the SluggableRepository with the isSlugUniqueFor-Method for checking against the database, and an additional isSlugUnique-Method which is used to check against other entities in the current changeset.

Testcases are provided. :)

### DIFF
--- a/src/EventSubscriber/SluggableSubscriber.php
+++ b/src/EventSubscriber/SluggableSubscriber.php
@@ -99,9 +99,14 @@ final class SluggableSubscriber implements EventSubscriber
 
         $uniqueSlug = $slug;
 
+        $sluggableRepository = $this->entityManager->getRepository(get_class($sluggable));
+        if (!$sluggableRepository instanceof SluggableRepositoryInterface) {
+            $sluggableRepository = $this->defaultSluggableRepository;
+        }
+
         while (! (
-            $this->defaultSluggableRepository->isSlugUniqueFor($sluggable, $uniqueSlug)
-            && $this->isSlugUniqueInUnitOfWork($sluggable, $uniqueSlug)
+            $sluggableRepository->isSlugUniqueFor($sluggable, $uniqueSlug)
+            && $this->isSlugUniqueInUnitOfWork($sluggableRepository, $sluggable, $uniqueSlug)
         )) {
             $uniqueSlug = $slug . '-' . ++$i;
         }
@@ -109,11 +114,11 @@ final class SluggableSubscriber implements EventSubscriber
         $sluggable->setSlug($uniqueSlug);
     }
 
-    private function isSlugUniqueInUnitOfWork(SluggableInterface $sluggable, string $uniqueSlug): bool
+    private function isSlugUniqueInUnitOfWork(SluggableRepositoryInterface $repository, SluggableInterface $sluggable, string $uniqueSlug): bool
     {
         $scheduledEntities = $this->getOtherScheduledEntities($sluggable);
         foreach ($scheduledEntities as $entity) {
-            if ($entity->getSlug() === $uniqueSlug) {
+            if (!$repository->isSlugUnique($uniqueSlug, $sluggable, $entity)) {
                 return false;
             }
         }
--- a/src/EventSubscriber/SluggableSubscriber.php
+++ b/src/EventSubscriber/SluggableSubscriber.php
@@ -11,7 +11,7 @@ use Doctrine\ORM\Event\LoadClassMetadataEventArgs;
 use Doctrine\ORM\Events;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Knp\DoctrineBehaviors\Contract\Entity\SluggableInterface;
-use Knp\DoctrineBehaviors\Repository\DefaultSluggableRepository;
+use Knp\DoctrineBehaviors\Contract\Repository\SluggableRepositoryInterface;
 
 final class SluggableSubscriber implements EventSubscriber
 {
@@ -21,7 +21,7 @@ final class SluggableSubscriber implements EventSubscriber
     private const SLUG = 'slug';
 
     /**
-     * @var DefaultSluggableRepository
+     * @var SluggableRepositoryInterface
      */
     private $defaultSluggableRepository;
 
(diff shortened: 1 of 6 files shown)


TomasVotruba:
Hi, I don't know the feature in gedmo.
What is it for?

I wonder if slug should be non-unique anytime. Thought?

crazy-weasel:
Hi!

You can think of it as a namespace for slugs. 

Example: 
Website promoting companies, each company can has serveral pages.
URL structure: `/[company-slug]/[page-slug]`

Assume every company has an about-page.

```
Company-Entity:
 - name
 - slug
```

```
Page-Entity:
 - name
 - slug
 - company
```

Without "unique_base":
`/acme/about`
`/contoso/about-1`

With page.company used as the base:
`/acme/about`
`/contoso/about`

src/EventSubscriber/SluggableSubscriber.php
```diff
@@ -99,21 +99,29 @@ private function generateUniqueSlugFor(SluggableInterface $sluggable): void
 
         $uniqueSlug = $slug;
 
+        $sluggableRepository = $this->entityManager->getRepository(get_class($sluggable));
```
TomasVotruba, line 102: Repository should not depend on repository.

Better use only `SluggableRepositoryInterface` interface to get the needed service.

src/Repository/DefaultSluggableRepository.php
```diff
@@ -40,4 +41,12 @@ public function isSlugUniqueFor(SluggableInterface $sluggable, string $uniqueSlu
...
             ->getSingleScalarResult();
     }
+
+    public function isSlugUnique(
+        string $uniqueSlug,
+        SluggableInterface $newOrUpdated,
+        SluggableInterface $exisiting
+    ): bool {
+        return $uniqueSlug !== $exisiting->getSlug();
+    }
```
TomasVotruba, line 51: The name `isSlugUnique` is suggesting there can be conflicting slugs, which sound like conflicting ids.

What does it actually do?

tests/ORM/SluggableWithUniqenessAndOwnRepositoryTest.php
```diff
@@ -0,0 +1,172 @@
...
+
+        $this->entityManager->persist($entity2);
+        $this->entityManager->flush();
+
+        $id1 = $entity1->getId();
+        $this->assertNotNull($id1);
+
+        $id2 = $entity2->getId();
+        $this->assertNotNull($id2);
+
```
TomasVotruba, line 119: These test are not needed, as we're not testing the doctrine persistance here. It makes test much less readable as that's Doctrine, not slugs.

tests/Fixtures/Entity/SluggableWithUniqenessAndOwnRepositoryEntity.php
```diff
@@ -0,0 +1,75 @@
...
+declare(strict_types=1);
+
+namespace Knp\DoctrineBehaviors\Tests\Fixtures\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+use Knp\DoctrineBehaviors\Contract\Entity\SluggableInterface;
+use Knp\DoctrineBehaviors\Model\Sluggable\SluggableTrait;
+
+/**
+ * @ORM\Entity(repositoryClass="Knp\DoctrineBehaviors\Tests\Fixtures\Repository\SluggableWithUniqenessAndOwnRepositoryRepository")
```
TomasVotruba, line 12: This should work without dependency on specific repository.

```suggestion
 * @ORM\Entity()
```

src/Repository/DefaultSluggableRepository.php
crazy-weasel: The method in the default-repository is just the extracted comparison of slugs from the [SluggableEventSubscriber::isSlugUniqueInUnitOfWork](https://github.com/KnpLabs/DoctrineBehaviors/blob/master/src/EventSubscriber/SluggableEventSubscriber.php#L116) method.

It checks, wether the slugs of two entities collide. I moved the check to the repository, because different entities can have different uniqueness constraints. Like, the default one with no constraints at all or my example with the pages per company.

I realize that the naming is bad, because it's just the comparison of two entities, and not the whole set of entities.

src/EventSubscriber/SluggableSubscriber.php
crazy-weasel: I get the repository for the entity, because different types of entities can have different requirements on the slug. See my comment on "isSlugUnique"

TomasVotruba: Will this work with https://github.com/KnpLabs/DoctrineBehaviors/pull/541#discussion_r439817970 ?

crazy-weasel: The SluggableEventSubscriber has a [depdendency on SluggableRepositoryInterface](https://github.com/KnpLabs/DoctrineBehaviors/blob/dc2c92532196f9d96d77880bc72a8aed67771f7b/src/EventSubscriber/SluggableSubscriber.php#L35). This repository [is used as fallback ](https://github.com/KnpLabs/DoctrineBehaviors/blob/dc2c92532196f9d96d77880bc72a8aed67771f7b/src/EventSubscriber/SluggableSubscriber.php#L103)if the repository from the entity does not implement SluggableRepositoryInterface.

TomasVotruba:
Closing for lack of feedback for last ~8 months

jkabat:
@TomasVotruba I want to use custom SluggableRepository but with current setup (subscriber depends on exact final class) it is not possible.

My use case is to use same SLUG for translation entities with different locale. UNIQUE constraint obviously does not work as SLUG is generated in prepersist event.

What do you suggest?

TomasVotruba:
Hi, I'm not sure as I don't use this specific behavior. It seems like extracting an interface would help

jkabat:
Thats what I think. Would you accept a PR? I would need a helping hand, as Im locked on PHP 7.4 and cant use master... would be possible to back merge?

TomasVotruba:
PHP 8 is needed to be able contribute, or you can try Docker

PR #544, opened 2020-03-31
Fix broken old bandle register class

### LINKED ISSUE #542: Release 2.0.2 BC breaks due to bundle namespace change
Bundle namespace changed from   
`Knp\DoctrineBehaviors\Bundle\DoctrineBehaviorsBundle` to   `Knp\DoctrineBehaviors\DoctrineBehaviorsBundle`  
So we need to update our `config/bundles.php`

Change comes from this PR https://github.com/KnpLabs/DoctrineBehaviors/pull/534 
Maybe reverting it and release a 2.0.3? 

I guess a lot of people will update their vendors soon, due to todays symfony security release 4.4.7 / 5.0.7.

uginroot:
closes #542

### DIFF
--- a/ecs.yaml
+++ b/ecs.yaml
@@ -35,6 +35,11 @@ parameters:
         SlevomatCodingStandard\Sniffs\Classes\UnusedPrivateElementsSniff:
             - "tests/Fixtures/Entity/SluggableWithoutRegenerateEntity.php"
 
+    exclude_files:
+        # Old bundle class name duplicate
+        - "src/Bundle/DoctrineBehaviorsBundle.php"
+        - "src/DoctrineBehaviorsBundle.php"
+
 services:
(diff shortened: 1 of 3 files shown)


uginroot:
ecs does not allow class names to be called by the same name, even in different namespaces, but in this case it must be done. What to do in this case?

uginroot:
I apologize for spam, the local version of ecs on windows works a little wrong

TomasVotruba:
No troubles. Could you squash it into 1 commit? I'll handle the cs

bobdenotter:
Thank you for looking into this, @uginroot and @TomasVotruba 🙏

TomasVotruba:
Thanks

TomasVotruba:
Just released v2.0.3 with this fix.
Thanks for the help!

PR #545, opened 2020-04-06
Package own EventSubscriberInterface 

wimwinterberg:
Created an EventSubscriberInterface and use it when registerForAutoconfiguration doctrine.event_subscriber instead of the using Doctrine\Common\EventSubscriber.

Using Doctrine\Common\EventSubscriber reulted in cache:clear for me to fail

### DIFF
--- a/src/Bundle/DependencyInjection/DoctrineBehaviorsExtension.php
+++ b/src/Bundle/DependencyInjection/DoctrineBehaviorsExtension.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace Knp\DoctrineBehaviors\Bundle\DependencyInjection;
 
-use Doctrine\Common\EventSubscriber;
+use Knp\DoctrineBehaviors\EventSubscriber\EventSubscriberInterface;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Extension\Extension;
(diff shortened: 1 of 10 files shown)


wimwinterberg:
To fix the ECS issues I have to also rename the subscriber classes. Any thoughts on that?

TomasVotruba:
> Using Doctrine\Common\EventSubscriber reulted in cache:clear for me to fail

How exactly?
This would break all the existing code

wimwinterberg:
I don't know exactly yet. But what happens is that my php memory will grow enormous and will finally crashes. When I activate xdebug, I see max_nesting_level reaching even 9999 during warmup.. Maybe it's because of other packages. I use vichupload and shapecode/cron-bundle from which I see also uses doctrine subscriber tag. My wild guess is that is has something to do with services being autowired unintentionally now.

wimwinterberg:
@TomasVotruba sorry for wasting your time. I managed to solve the memory issue by rewriting some of my classes which also implements the EventSubscriber by using Service Locator now instead of requiring some of the used services. Now there are no recursive issues anymore.

TomasVotruba:
No time wasted :). Could you share more practical code (diff) so others know how to solve this if they have similar problem too?

wimwinterberg:
@TomasVotruba 

A diff wont help but maybe this small comment in the readme or somewhere else might.


"
When you run into memory issues during cache:clear or warmup. You might fix the problem to check your EventSubscribers which have some dependencies on other services and use a locator for those services  as described in https://symfony.com/doc/current/service_container/service_subscribers_locators.html
"

TomasVotruba:
I never used service subscribers, so not sure what it means.

wimwinterberg:
@TomasVotruba do you want me to do something additional to explain this more ? 
Let me know.

TomasVotruba:
I don't have such problem, so I'm ok :+1:

PR #546, opened 2020-04-06
Bugfix for Tree: getRealMaterializedPath

prof-web-coding:
See Added Test

### DIFF
--- a/src/Model/Tree/TreeNodeMethodsTrait.php
+++ b/src/Model/Tree/TreeNodeMethodsTrait.php
@@ -28,6 +28,9 @@ trait TreeNodeMethodsTrait
 
     public function getRealMaterializedPath(): string
     {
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Hi, thanks for fix + test. I don't use Tree at all, so I can't read the test.

Could you explain in human words what was the issue?

prof-web-coding:
If I do ```setChildNodeOf(null);``` it should result in a materialized path of consisting of just the "MaterializedPathSeparator" (e.g. '/'). This was fine. If I then add a child to this node, the materialized path of of this child should be "MaterializedPathSeparator" plus the id of the parent node (e.g. '/1'). Without the fix, it was twice the "MaterializedPathSeparator" plus the id. (e.g. '//1');

prof-web-coding:
Do you need more info @TomasVotruba ?

TomasVotruba:
I see. All good, thanks

PR #554, opened 2020-04-27
Feature: Added support for custom timestampable fieldnames

AndreasA:


src/EventSubscriber/TimestampableSubscriber.php
```diff
@@ -58,6 +64,25 @@ public function getSubscribedEvents(): array
         return [Events::loadClassMetadata];
     }
 
+    /**
+     * @return string[]
+     */
+    private function getTimestampableProperties(string $className): array
+    {
+        $properties = [];
+        foreach (['getCreatedAtProperties', 'getUpdatedAtProperties'] as $method) {
```
TomasVotruba, line 73: Instead of magic, here can be `is_a()`

src/Contract/Entity/TimestampableInterface.php
```diff
@@ -8,13 +8,27 @@
...
+     * @return string[]
+     */
+    public static function getCreatedAtProperties(): array;
+
+    /**
+     * Returns an array of updatedAt properties.
+     *
+     * @return string[]
+     */
+    public static function getUpdatedAtProperties(): array;
```
TomasVotruba, line 33: How could we get rid of static? These method are available on entity object

AndreasA: I need it in the event subscriber which is why they are static as I don't have an object there. If you have a better idea though...

TomasVotruba: I'd get object first somehow and work with that.

AndreasA: I will think about a better solution. what about an annotation on the properties instead?

AndreasA: @TomasVotruba I have now made them non-static.
I thought about using annotations but the issue then is, I also would need to access the data inside the class itself and I didn't want to read annotations inside the entity, so I made them non static.

However, Doctrine has already a method to create an instance of the entity (in the classmetadata) without executing the constructor, so it was easy to create an instance. If not, I could also have used reflection but if Doctrine already has a method, I think it is better to use that one (e.g. caching etc.).

src/EventSubscriber/TimestampableSubscriber.php
AndreasA: @TomasVotruba modified this whendoing the change below.

src/Model/Timestampable/TimestampableCurrentDateTimeMethodTrait.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Returns a DateTimeInterface object with the current timestamp.
+     *
+     * @throws ShouldNotHappenException
+     */
+    public function getCurrentDateTime(): DateTimeInterface
+    {
+        // Create a datetime with microseconds.
+        $dateTime = DateTime::createFromFormat('U.u', sprintf('%.6F', microtime(true)));
```
AndreasA, line 22: @TomasVotruba Btw. I have been wondering about this code. Why is it so complicated.
A PHP DateTime object contains microseconds if I create it like  `new DateTime()`, is already assigned the default timezone and the current datetime (with microseconds). At least since PHP 7.1 and required PHP is 7.2 or higher.

AndreasA:
@TomasVotruba  Do you need any further changes, so this can be merged? For SoftDetelte I would create a separate MR, once this is merged.

src/Contract/Entity/TimestampableInterface.php
```diff
@@ -8,13 +8,27 @@
...
+     * @return string[]
+     */
+    public function getCreatedAtProperties(): array;
+
+    /**
+     * Returns an array of updatedAt properties.
+     *
+     * @return string[]
+     */
+    public function getUpdatedAtProperties(): array;
```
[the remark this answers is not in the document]
AndreasA: In theory the changes to the traits could also break older installations, especially if one did not use the master trait but a sub-trait.

I could create a whole namespace for these files in `Timestampable` that are unrelated to the previous ones? Though that means duplicate implementations. I could mark the old one as deprecated,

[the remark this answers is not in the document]
TomasVotruba: This is about interfaces, not traits.

If `TimestampableInterface` has a new method, all the classes that implement it will break.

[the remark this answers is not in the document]
AndreasA: As long as those classes used the Trait it will not break.

If the classes used some of the traits it will still break, even if I create a separate Interface.

So just creating another Interface will not fully fix BC compatibility.

However, it will decrease the chance of it breaking, so I can do that.

Just wanted to make sure we are clear that it will not fully fix BC.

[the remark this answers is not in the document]
TomasVotruba: > As long as those classes used the Trait it will not break.

That's an assumption. Only the interface is reliable source, everything else can be custom.
Any change of interface will break the contract. We can allow that in 3.0, but it's too soon. 2.0 was released just 6 months ago.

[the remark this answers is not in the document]
AndreasA: Yes it is an assumption which is why I said just creating another interface will not be 100% backwards compatible either 
as one could have used one of the traits but modified one of the original methods which will not work, if I just create another interface.

I am quite sure it will break just as much as just coying the interface as traits are external usable files.

The only 100% solution would be to create a separate interface and separate traits. 

For isntance `TimestampableMethodsTrait` uses completely different methods than before.

Due to the different methods I will need to create separate traits anyway I think.

I have no issue with keeping the original as is and create subfolders in the timestampable parts like `Configurable` and the new traits / itnerfaces there.

That is the only way we can be sure regarding backwards compatibility and if you are fine with that I will change it accordingly. 

However, I do not want to just create a separate interface and still break backwards compatibility because people expect the traits to works as before. I just want to make sure you are fine with creating separate interfaces and traits.

TomasVotruba:
Closing as not adjusted to avoid BC breaks

AndreasA:
@TomasVotruba Sorry, I totally forgot I still had this one open. However, I currently don't have time to fix it. If I find the time, I will create a new PR and add in the fixes regarding BC.

TomasVotruba:
No problem, I forgot about it too :D thanks for reaching out! Any furute PR is of course welcomed :+1:

PR #555, opened 2020-04-27
Avoid database access loading metadata

### LINKED ISSUE #553: [Timestampable] Unnecessary database access
Hi,

In [TimestampableSubscriber](https://github.com/KnpLabs/DoctrineBehaviors/blob/a076c4a4794d05aeaac81ec9ae18009d1696a939/src/EventSubscriber/TimestampableSubscriber.php#L63) a call to database connection is done to choose between `datetimetz` and `datetime` metadata format to add.
It throws an exception if you install your vendors before starting the database (or if you don't need a started database in the CI for example). Maybe it would be better to set the format in the configuration with a defaut value?

If it's ok for you, I will create the corresponding PR.

pmontoya:
To fix #553
Move datetime type field from database type detection to service configuration

### DIFF
--- a/tests/ORM/TimestampableMakeEntityTest.php
+++ b/tests/ORM/TimestampableMakeEntityTest.php
@@ -29,9 +29,4 @@ final class TimestampableMakeEntityTest extends AbstractBehaviorTestCase
         $doctrineEventManager->dispatchEvent(Events::loadClassMetadata, $eventArgs);
         $this->expectNotToPerformAssertions();
     }
-
-    protected function provideCustomConfig(): ?string
-    {
-        return __DIR__ . '/../config/config_test_with_timestampable_entity.yaml';
-    }
 }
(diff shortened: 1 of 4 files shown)


tests/ORM/TimestampableMakeEntityTest.php
```diff
@@ -29,9 +29,4 @@ public function testMakeEntityEmptyEvent(): void
         $doctrineEventManager->dispatchEvent(Events::loadClassMetadata, $eventArgs);
         $this->expectNotToPerformAssertions();
     }
-
-    protected function provideCustomConfig(): ?string
```
pmontoya, line 33: I think it's no more necessary. What do you think?

TomasVotruba:
Looks good!

TomasVotruba:
Thank you

PR #557, opened 2020-04-29
[Configuration] Respect symfony recommendation on bundle configuration

### LINKED ISSUE #556: [Improvement] Configuration tree parse
Hi,

Do you think it would be better to have a specific configuration file (parsed thanks to a ConfigurationInterface) to set parameters than overriding parameters defined in services.yaml?

The symfony documentation : https://symfony.com/doc/current/bundles/configuration.html

We should then create configuration for composer flex plugin.

I would be happy to create the PR if it's ok for you but it will create a BC break.

pmontoya:
Fixes #556 

I tried to respect symfony recommendation on bundle configuration as explained here : https://symfony.com/doc/current/bundles/configuration.html

I switched services from yaml format to xml has recommanded by symfony for distributed bundle and make them private (as recommanded too).

I added the configuration documentation.

I didn't update the CHANGELOG.md file


src/Bundle/DependencyInjection/DoctrineBehaviorsExtension.php
```diff
@@ -4,29 +4,44 @@
...
-
     /**
      * @param string[] $configs
      */
     public function load(array $configs, ContainerBuilder $containerBuilder): void
     {
-        $loader = new YamlFileLoader($containerBuilder, new FileLocator(__DIR__ . '/../../../config'));
-        $loader->load('services.yaml');
+        $loader = new XmlFileLoader($containerBuilder, new FileLocator(dirname(__DIR__) . '/../Resources/config'));
+        $loader->load('services.xml');
```
TomasVotruba, line 20: Keep services in `services.yaml`. That's unrlated change and I want to stick with YAML

pmontoya: Ok. I'll do this. Just to understand : why ? ^^

TomasVotruba: Yaml is more friendly to new devs, as it's default in Symfony docs and StackOverflow answers.

pmontoya: Ok for application and config override but for internal bundle config, XML is the standard, no?

README.md
```diff
@@ -22,6 +22,10 @@ It currently handles:
 composer require knplabs/doctrine-behaviors
 ```
 
+##Configure
```
Kocal, line 25:
```suggestion
## Configure
```

docs/blameable.md
```diff
@@ -54,7 +54,9 @@ var_dump($updatedBy);
 By default, no user entity is provided. You need to specify the User class with a new parameter in your config:
 
 ```yaml
-# services.yaml
-parameters:
-    doctrine_behaviors_blameable_user_entity: App\Entity\User
+# doctrine_behaviors.yaml
+doctrine_behaviors:
```
Kocal, line 58: What about this?
```suggestion
# knp_doctrine_behaviors.yaml
knp_doctrine_behaviors:
```

KnpLabs's packages like [KnpSnappyBundle](https://github.com/KnpLabs/KnpSnappyBundle#configuration) or [KnpPaginatorBundle](https://github.com/KnpLabs/KnpPaginatorBundle#configuration-example) prefix their config file and root node's name by `knp_`

Kocal: Also I think it requires to update the DoctrineBehaviors bundle name to  `KnpDoctrineBehaviorsBundle` I think, see https://symfony.com/doc/current/bundles/configuration.html#using-the-bundle-extension

docs/config.md
```diff
@@ -0,0 +1,14 @@
+#Configuration reference
```
Kocal, line 1:
```suggestion
# Configuration reference
```

docs/config.md
```diff
@@ -0,0 +1,14 @@
...
+
+The full configuration with default values
+
+```yaml
+doctrine_behaviors:
```
Kocal, line 6:
```suggestion
knp_doctrine_behaviors:
```

docs/timestampable.md
```diff
@@ -46,7 +46,9 @@ var_dump($category->getUpdatedAt());
 By default, timestampable use `datetime` field type. If you want use `datetimetz`, just modify parameters in your config:
 
 ```yaml
-# services.yaml
-parameters:
-    doctrine_behaviors_timestampable_date_field_type: "datetimetz"
+# doctrine_behaviors.yaml
+doctrine_behaviors:
```
Kocal, line 50:
```suggestion
# knp_doctrine_behaviors.yaml
knp_doctrine_behaviors:
```

docs/translatable.md
```diff
@@ -222,8 +222,10 @@ class SomeClass
 By default, translation relations are lazy loaded. To change that, just modify parameters in your config:
 
 ```yaml
-# services.yaml
-parameters:
-    doctrine_behaviors_translatable_fetch_mode: "LAZY"
-    doctrine_behaviors_translation_fetch_mode: "LAZY"
+# doctrine_behaviors.yaml
+doctrine_behaviors:
```
Kocal, line 226:
```suggestion
# knp_doctrine_behaviors.yaml
knp_doctrine_behaviors:
```

src/Bundle/DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,63 @@
...
+namespace Knp\DoctrineBehaviors\Bundle\DependencyInjection;
+
+use Symfony\Component\Config\Definition\Builder\TreeBuilder;
+use Symfony\Component\Config\Definition\ConfigurationInterface;
+
+final class Configuration implements ConfigurationInterface
+{
+    public function getConfigTreeBuilder()
+    {
+        $treeBuilder = new TreeBuilder('doctrine_behaviors');
```
Kocal, line 14:
```suggestion
        $treeBuilder = new TreeBuilder('knp_doctrine_behaviors');
```

src/Bundle/DependencyInjection/DoctrineBehaviorsExtension.php
pmontoya: Done

docs/blameable.md
pmontoya: That's why I let doctrine_behaviors (I prefer prefixing with knp too). 
So I create a new Bundle, deprecate the previous one and force the name.

README.md
pmontoya: Good catch. Thank you

docs/config.md
pmontoya: Good catch. Thank you

docs/config.md
pmontoya: Done

docs/timestampable.md
pmontoya: Done

docs/translatable.md
pmontoya: Done

src/Bundle/DependencyInjection/Configuration.php
pmontoya: Done

src/Resources/config/services.yaml
```diff
@@ -0,0 +1,80 @@
...
+            - '@doctrine.orm.entity_manager'
+
+    knp_doctrine_behaviors.event_subscriber.blameable:
+        class: Knp\DoctrineBehaviors\EventSubscriber\BlameableSubscriber
+        arguments:
+            - '@knp_doctrine_behaviors.user_provider'
+            - '@doctrine.orm.entity_manager'
+            - ~ # Replaced by knp_doctrine_behaviors.blameable.user_entity
+        tags:
+            - 'doctrine.event_subscriber'
```
TomasVotruba, line 34: This tagging is better handled in the extension by type.

src/Resources/config/services.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+    knp_doctrine_behaviors.event_subscriber.tree:
+        class: Knp\DoctrineBehaviors\EventSubscriber\TreeSubscriber
+        tags:
+            - 'doctrine.event_subscriber'
+
+    knp_doctrine_behaviors.event_subscriber.uuidable:
+        class: Knp\DoctrineBehaviors\EventSubscriber\UuidableSubscriber
+        tags:
+            - 'doctrine.event_subscriber'
```
TomasVotruba, line 80: Why are these services needed to be added manually? All of them can be autodiscovered.

docs/blameable.md
TomasVotruba: I think `octrine_behaviors` is unique enough now. No need to change it

TomasVotruba:
I'm kinda lost what is purpose of this PR. It seems just changing YAML syntax to another YAML syntax. Just with more complex code that is very hard to read.

What is the original issue you have?

pmontoya:
@TomasVotruba : I'm ok with you, your code was more simple than v1.x and that's a good choice.
I haven't issue with your code.

I create this PR to allow to configure the bundle as recommended by Symfony for reusable bundles. It will allow to create a recipe for flex.

For that reason all services became private and parameters are injected in the DependencyInjection replacing missing arguments (that why I have to tag manually).
I dont want to add complexity to the bundle, just to respect Symfony standards to allow other contributors to easily contribute to great project.

webda2l:
Disagree with the commit that change XML to YAML, XML still is, even with recent bundles (by example https://github.com/symfony/maker-bundle/blob/master/src/Resources/config/services.xml) the recommended way.. until SF6 https://github.com/symfony/symfony/issues/37186
Otherwise, agree with the two others commits that go back with Symfony recommendations, but required another major version for BC.. :/

pmontoya:
@webda2l : I would that to switch to php as recommended now for SF6 but as the PR is not approved for other reason, I haven't take the time to update my PR.

@TomasVotruba : Do you think this PR as a chance to be merged? If yes, I switch configuration to php, if not, I'll close it.

TomasVotruba:
I'm open to switching YAML to PHP in standalone PR. 
Without switching back to old-bundle-extension-configuration approach.

webda2l:
@pmontoya sure, otherwise looks good to me :)
@TomasVotruba has done a hard work with this bundle and if he doesn't want add back the configuration part in the SF way, maybe he is ok if you create your own SF bundle that depends of this one, to add missing parts you have done here ?
Better than nothing and so.. allow reusable bundles, custom configuration, create a recipe for flex, .. 
https://symfony.com/doc/current/bundles/prepend_extension.html

TomasVotruba:
Bundles and extension have less and less support in Symfony. ENV, parameters and Flex are way to go.

Recommeneded reading: [How to Manage Configuration in Symfony without Bundle, Extension, and Configuration?](https://www.tomasvotruba.com/blog/2018/11/29/how-to-manage-configuration-in-symfony-without-bundle-extension-and-configuraiton/)

As you can see in https://github.com/symfony/symfony/issues/37186, the similar path will follow for internal *bundles* too.

TomasVotruba:
Closing as scope narrowed in: https://github.com/KnpLabs/DoctrineBehaviors/pull/557#issuecomment-644793764

webda2l:
@TomasVotruba Like most of your blog post but not fully convinced by this one. And Symfony ride of XML services definitions, not ride of Configuration & DependencyIjection\***Extension from what I know..
Having https://github.com/KnpLabs/DoctrineBehaviors/blob/master/composer.json#L28-L29 in ```require``` part whereas it should be in ```suggest``` is disappointing.
With the right Configuration, depending of enabled features, user should be warned to require theses deps. I'm not the first one to say that.
You can do it currently but this is the goal of 
https://github.com/KnpLabs/DoctrineBehaviors/pull/557/files#diff-bafb72d16cd1976e3fd2145236904e83 & https://github.com/KnpLabs/DoctrineBehaviors/pull/557/files#diff-b17ec04a6d8762db9db37e6db30d42cf which have their usefulness.

pmontoya:
@TomasVotruba : I'm fully agree with @webda2l last comment and I strongly prefer contributing to this very good bundle than creating a new one, depending of this one. 
Never mind...

TomasVotruba:
@webda2l It's actually not an idea from my head, but solution based on Symfony-less approach proclaimed officialy as default since Symfony 4.
Here is it explained nicely in symfony/flex recipe: https://stackoverflow.com/a/58595941/1348344

Also extension-scoped parameters have to be handled manully per services, while parameters can be re-used and autowired automatically. This allows to use [bind params](https://symfony.com/blog/new-in-symfony-4-1-autowiring-improvements) (Symfony 4.1) and use PSR-4 autodiscovery.

PR #561, opened 2020-05-09
Feature: Added support for hashids

tsdevelopment:
Instead of using Uuids sometimes it's nicer to use hashIds (known from Youtube video IDs).


tests/ORM/HashidableTest.php
```diff
@@ -0,0 +1,33 @@
...
+
+        $this->entityManager->persist($entity);
+        $this->entityManager->flush();
+
+        $id = $entity->getId();
+        $this->assertNotNull($id);
+
+        $hashId = $entity->getHashId();
+        $this->assertNotNull($hashId);
+        $this->assertNotEmpty($hashId);
```
TomasVotruba, line 29: The hash is value should be asserted more precisely than just not `null`. Just to be sure the hashable works, and it's not just generated value by Doctrine.

tsdevelopment: @TomasVotruba I added two more tests. Currently some checks from this repository doesn't work. But I think it's b/c they also don't work on the master branch. Locally everything works fine.

TomasVotruba:
Thanks :+1:

bobdenotter:
Hi @tsdevelopment,

I'm getting the following error: 

```
!!    Cannot autowire service "Knp\DoctrineBehaviors\EventSubscriber\HashidableEv
!!    entSubscriber": argument "$hashids" of method "__construct()" references cl
!!    ass "Roukmoute\HashidsBundle\Hashids" but no such service exists. Try chang
!!    ing the type-hint to "Hashids\HashidsInterface" instead.
```

At first glance, this can be fixed by modifying that constructor in `HashidableEventSubscriber`:

```php
public function __construct(EntityManagerInterface $entityManager, HashidsInterface $hashids)
```

jkabat:
What if we do not want to have hard dependency on HashIds library because we do not use it? Should it be part of the core or not?

TomasVotruba:
@jkabat It's design choice to prevent maintaining 12 repositories.

PR #570, opened 2020-06-16
Removed hardcoded identifier.

ONiAd:
Use classmetadatainfo to de-hardcode 'id' (for other identifier names like uuid). Allows multiple identifiers too.


TomasVotruba:
Could you add test that failed before this fix? Just so we're sure it still works

ONiAd:
> Could you add test that failed before this fix? Just so we're sure it works

Sorry, don't know how. It's running well on my local though :|

TomasVotruba:
Basically copy this: https://github.com/KnpLabs/DoctrineBehaviors/blob/3eed95ec2e2512b8995ce039d0767bac661fd7c4/tests/ORM/TranslatableTest.php#L30-L50

And instead of `TranslatableEntity` create own class, that has different id name.

TomasVotruba:

Thank for understanding.

PR #577, opened 2020-06-21
[Hashids] Enable optionaly if Roukmoute bundle is registered

TomasVotruba:


TomasVotruba:
No. That would extend install step by 1 package per behavior and create eventually 10 extra packages. To much code to maintain.

Clear dependency in `composer.json` makes usage less WTF, see https://matthiasnoback.nl/2014/04/theres-no-such-thing-as-an-optional-dependency/

TomasVotruba:
See https://github.com/KnpLabs/DoctrineBehaviors/commit/03c4474b1a3b318cdacb0f8ae7188262be7b2831, I forgot to add it to this PR

jordisala1991:
> Clear dependency in composer.json makes usage less WTF, see https://matthiasnoback.nl/2014/04/theres-no-such-thing-as-an-optional-dependency/

It is strange to be forced to have the roukmoute bundle (that comes with the framework extra bundle) installed if I don't want to use that bundle or the hashid feature. That's why composer have the suggest feature.

TomasVotruba:
Personally I don't read suggest at all. I believe package all it needs when I install it. Otherwise what's the point of `require` section?

jordisala1991:
IMO the point of require section is to give you the minimum that is required to use the package (won't break if there is something missing). If there are optional features, that are only enabled under certain conditions, and there is not risk of breaking the app if missing, that's not a required dependency but an optional one.

TomasVotruba:
I'd only repeat what I already shared: https://matthiasnoback.nl/2014/04/theres-no-such-thing-as-an-optional-dependency/

PR #602, opened 2020-10-26
Move package-builder to dev dependency

maxhelias:
symplify/package-builder does not appear to be used other than for testing purposes


TomasVotruba:
Good catch :+1:

PR #603, opened 2020-10-27
TreeNodeMethodsTrait must be compatible with TreeNodeInterface

maxhelias:


TomasVotruba:
Could you add failing test to show the bug you have before this changes? So we know we have it covered and it won't happen again

maxhelias:
In fact, we don't have a php error but PhpStorm raise a warning
Capture d’écran 2020-10-27 à 13 41 48

TomasVotruba:
It might be bug in PHPStorm resolution then.

The `self` in interface = the interface itself, so PHP code is correct here

maxhelias:
Yes and no, as they are a php trait we would need more a way to link it with this interface. I did'nt find anything on the phpdoc level.

TomasVotruba:

Thank for understanding.

PR #607, opened 2020-11-06
feature/fix-bug-with-mapped-superclass

skrut:
- fixed bug with MappedSuperclass

If class has `MappedSuperclass` annotation and inherited from `class  implements TimestampableInterface` - we will got the following error:
`Duplicate definition of column 'created_at' on entity in a field or discriminator column mapping.` The same for updatedAt field. Pls have a look to fix.

### DIFF
--- a/ecs.php
+++ b/ecs.php
@@ -40,7 +40,6 @@ return static function (ContainerConfigurator $containerConfigurator): void {
         PropertyTypeHintSniff::class => null,
         PhpUnitStrictFixer::class => ['tests/ORM/TimestampableTest.php'],
         UnusedPrivateElementsSniff::class => ['tests/Fixtures/Entity/SluggableWithoutRegenerateEntity.php'],
-        FinalClassFixer::class => ['tests/Fixtures/Entity/TimestampableMappedSuperclassEntity.php'],
         OrderedImportsFixer::class => ['tests/Fixtures/Entity/TimestampableMappedSuperclassEntity.php'],
     ]);
 
--- a/ecs.php
+++ b/ecs.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 use PhpCsFixer\Fixer\ClassNotation\FinalClassFixer;
 use PhpCsFixer\Fixer\Comment\HeaderCommentFixer;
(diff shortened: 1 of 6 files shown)


Kocal:
Is it possible to add a test for prevent any regressions?

skrut:
> Is it possible to add a test for prevent any regressions?

Yes, I will do that asap.

ecs.php
```diff
@@ -39,6 +40,8 @@
         PropertyTypeHintSniff::class => null,
         PhpUnitStrictFixer::class => ['tests/ORM/TimestampableTest.php'],
         UnusedPrivateElementsSniff::class => ['tests/Fixtures/Entity/SluggableWithoutRegenerateEntity.php'],
+        FinalClassFixer::class => ['tests/Fixtures/Entity/TimestampableMappedSuperclassEntity.php'],
```
TomasVotruba, line 43: The class should be `abstract` and this line removed

ecs.php
```diff
@@ -39,6 +40,8 @@
...
+        OrderedImportsFixer::class => ['tests/Fixtures/Entity/TimestampableMappedSuperclassEntity.php'],
```
TomasVotruba, line 44: Instead of ignoring the error, let cs fix it.

```bash
composer fix-cs
```

skrut: ecs suggests to remove `use DateTimeInterface` but it uses in `@var DateTimeInterface`, after that phpstan will be unhappy.

ecs.php
skrut: Done.

TomasVotruba (approved):

ecs.php
TomasVotruba: I see. There is bug in one them, I'll look into that later, it's unrelated to this PR

TomasVotruba:
Thank you

roman-eonx:
Thanks, @TomasVotruba. Would you release a new version with this fix?

TomasVotruba:
Sure :+1: 

There you go https://github.com/KnpLabs/DoctrineBehaviors/releases/tag/v2.0.8

PR #611, opened 2021-01-07
Fallback to default locale if country locale fallback returns empty result

### LINKED ISSUE #562: Locale Fallback behaviour
In d9dde9789f4cbdfceb261c0d2c348f1b66fa0d08 the fallback behaviour was changed: 

**It was**
- locale (e.g. de-DE)
- fallback locale (e.g. de)
- default locale (e.g. en)

**Now it is**
For "complex locales" like "de-DE"
- locale (e.g. de-DE)
- fallback locale (e.g. de)

For "simple locales" like "de"
- locale (e.g. de)
- default locale (e.g. en)

This seems a little weird, because the default locale is now ignored in some cases. Why was this change made? 

Many browsers send Accept-Language headers with language and region like de-DE, for those users the new behaviour now only works if a page is available in de-DE or de. A different default locale is never provided. 

Do you have any recommendations, how the "old" behaviour could be restored?

Thank you very much :)

ossinkine:
Fixes #562 

Actually partially reverted d9dde97 and added the test.

### DIFF
--- a/src/Model/Translatable/TranslatableMethodsTrait.php
+++ b/src/Model/Translatable/TranslatableMethodsTrait.php
@@ -233,10 +233,13 @@ trait TranslatableMethodsTrait
     {
         $fallbackLocale = $this->computeFallbackLocale($locale);
 
-        if ($fallbackLocale === null) {
-            return $this->findTranslationByLocale($this->getDefaultLocale(), false);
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Thank you :+1:

PR #618, opened 2021-03-08
bump min version to PHP 7.4

TomasVotruba:


TomasVotruba:
Types on properties might cause some troubles, so we'll need a better community feedback on this change first.

Also replaced by PHP 8.0 bump

Issue #559, opened 2020-05-01
[Sluggable] Unable to provide locale for the slugger service


snikolow:
### Problem

I am playing around with the Sluggable behavior and it came to my attention that I am unable to provide a locale /even the fallback is not an option/ whenever I generate an entity slug.

Take a look at the following line - https://github.com/KnpLabs/DoctrineBehaviors/blob/v2.0.3/src/Model/Sluggable/SluggableMethodsTrait.php#L76 for the exact implementation and you can see that an option to provide locale is never considered. Even more, the method itself is marked as `private`, so not only we have to override the `generateSlug` method, but we also need the logic from `generateSlugValue`.

I've also took a look at the test cases and I believe they should be extended with some more complex localized strings for most languages. I am currently using the behaviour with cyrillic.

### Proposal

Move the class initialization to a new, `protected` method, something like `createSlugger` with nullable argument for `$locale` and then replace it by calling the new method in `generateSlugValue` so that we can override it in our entities and provide the locale on our own.

I am open to suggestions or further discussion, if a PR is necessary.

TomasVotruba:
Hi, I'm going through old opended issues and PRs to cleanup too-long-opened items and close them. Thank you for your contribution and time.

We needs a feature with test and passing CI to be able to merge it without letting a bug in.
If this is still relevant for you, please add code with test and make CI pass first. Then we'll able to merge your feature without further waiting :+1: 

Thank for understanding.

PR #620, opened 2021-03-18
Sluggable. Use default locale when generating entity slug

epashkov:
Hi there!

Relates issue: https://github.com/KnpLabs/DoctrineBehaviors/issues/559

I have a little difficulty with these PR. I do not know how to run the tests correctly and so I can not fix them.

Could somebody please help me?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,9 @@
         "symfony/translation-contracts": "^2.3",
         "nette/utils": "^3.2",
         "ramsey/uuid": "^4.0",
-        "symfony/polyfill-php80": "^1.22"
+        "symfony/polyfill-php80": "^1.22",
+        "symfony/property-access": "^5.2",
+        "symfony/yaml": "^5.2"
     },
     "require-dev": {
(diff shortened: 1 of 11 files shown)


epashkov:
Also we need slug entity field to be nullable. I use Sonata admin and there is related issue https://github.com/sonata-project/SonataAdminBundle/issues/5732

> Sonata uses Symfony forms bound to entities directly, even for the create form, that means that every field in your entity has to > have a nullable getter, or a default value that is not null.

epashkov:
How to restart deployment checks?

I fix tests, but I see that The operation was canceled.

TomasVotruba:
Hi, just push another commit. I've triggered it manually, but most jobs are failing

epashkov:
@TomasVotruba thanks. All tests are passed now, but static analysis is not. And I still don't understand what is wrong.

epashkov:
It seems that CI not working properly. I can run all checks on my local machine and no errors there.

PR #623, opened 2021-04-04
Use DateTimeInterface instead of DateTime

irmantas:
DateTimeInterface should be used instead of class DateTime.


TomasVotruba:

Thank for understanding.

Issue #558, opened 2020-05-01
Using Timestampable throws an error when creating a new entity


parijke:
I changed the entity like described in @TomasVotruba  blog https://www.tomasvotruba.com/blog/2020/01/06/migrate-gedmo-to-knplabs/ thinking it was very easy.

I can change an existing record and it captures the updatedAt time. But when creating a new record, an error is shown.

Can anyone help me with this?

```
TypeError:
Return value of App\Entity\Product::getCreatedAt() must implement interface DateTimeInterface, null returned

  at vendor/knplabs/doctrine-behaviors/src/Model/Timestampable/TimestampableMethodsTrait.php:16
  at App\Entity\Product->getCreatedAt()
     (vendor/symfony/property-access/PropertyAccessor.php:384)
  at Symfony\Component\PropertyAccess\PropertyAccessor->readProperty()
     (vendor/symfony/property-access/PropertyAccessor.php:93)
  at Symfony\Component\PropertyAccess\PropertyAccessor->getValue()
     (vendor/symfony/form/Extension/Core/DataMapper/PropertyPathMapper.php:49)
  at Symfony\Component\Form\Extension\Core\DataMapper\PropertyPathMapper->mapDataToForms()
     (vendor/symfony/form/Form.php:377)
  at Symfony\Component\Form\Form->setData()
     (vendor/symfony/form/Form.php:481)
  at Symfony\Component\Form\Form->initialize()
     (vendor/symfony/form/FormBuilder.php:206)
  at Symfony\Component\Form\FormBuilder->getForm()
     (vendor/easycorp/easyadmin-bundle/src/Controller/AdminControllerTrait.php:842)
  at EasyCorp\Bundle\EasyAdminBundle\Controller\EasyAdminController->createEntityForm()
     (vendor/easycorp/easyadmin-bundle/src/Controller/AdminControllerTrait.php:778)
  at EasyCorp\Bundle\EasyAdminBundle\Controller\EasyAdminController->createNewForm()
     (vendor/easycorp/easyadmin-bundle/src/Controller/AdminControllerTrait.php:908)
  at EasyCorp\Bundle\EasyAdminBundle\Controller\EasyAdminController->executeDynamicMethod()
     (vendor/easycorp/easyadmin-bundle/src/Controller/AdminControllerTrait.php:301)
  at EasyCorp\Bundle\EasyAdminBundle\Controller\EasyAdminController->newAction()
[...]

PR #637, opened 2021-05-25
Allow getCreatedAt and getUpdatedAt to return null

grizzlylab:
To fix this issue : https://github.com/KnpLabs/DoctrineBehaviors/issues/558

### DIFF
--- a/ecs.php
+++ b/ecs.php
@@ -37,7 +37,7 @@ return static function (ContainerConfigurator $containerConfigurator): void {
     ]);
 
     $services = $containerConfigurator->services();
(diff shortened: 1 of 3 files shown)


TomasVotruba:
Hi, thanks for the PR! Looks good :+1: 

> The only thing I found to make the pipeline pass is to disable the FinalClassFixer which is probably not a good thing.

This rule is now deprecated, so it can be removed from `ecs.php`
We handle this use case by Rector in better way.

TomasVotruba:
Looks good to me, thank you 😊

micronax:
Time for a new release? :)

PR #639, opened 2021-05-29
Handle embeddable as primary key in sluggables

frankverhoeven:
Also fixes tables with composite primary keys, and removes the assumption that `Sluggable::getId()` exists.


TomasVotruba:
Thank you for the contribution :+1:

PR #640, opened 2021-07-19
Make Sluggable getter and setter nullable

arnaud-ritti:
It resolve the bug `Expected argument of type "string", "null" given at property path "slug".` when you use Sonata Admin.


TomasVotruba:
Hi, thanks for contribution. We'll need a test case to cover this too.

TomasVotruba:
Closing for lack of feedback. We're narrowing issues to keep focus on active and engaging contributor to keep project growing. Thank you for understanding.

Feel free to re-open with failing test case.

PR #642, opened 2021-08-03
Sluggable + Translation - Allow getters to return null

patrick-benoit:
Allow new/empty translatable and sluggable objects to be used in forms, without throwing errors.

The error displayed was :
`Return value of My\Projet\Entity\EntityTranslation::getLocale() must be of the type string, null returned`

We noticed  #640 after making this commit, so we can remove all modifications to Sluggable to avoid any duplicate if needed.

### DIFF
--- a/src/Contract/Entity/SluggableInterface.php
+++ b/src/Contract/Entity/SluggableInterface.php
@@ -13,9 +13,9 @@ interface SluggableInterface
      */
     public function getSluggableFields(): array;
 
-    public function setSlug(string $slug): void;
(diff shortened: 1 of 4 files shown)


TomasVotruba:
Hi, thanks for contribution. We'll need a test case to verify this change works in the future too.

TomasVotruba:
Closing for lack of feedback. We're narrowing issues to keep focus on active and engaging contributor to keep project growing. Thank you for understanding.

Feel free to re-open with failing test case.

PR #635, opened 2021-05-05
Update .gitattributes file

ker0x:
These folders and files are not needed when fetching the package through Composer.

However, i'm not sure about the **utils** folder and **phpstan-extension.neon**.

PR #651, opened 2021-08-09
Update .gitattributes file

ker0x:
@TomasVotruba as suggested in #635, I opened a new PR that is sync with master

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,7 +1,14 @@
-.gitattributes     export-ignore
-.gitignore         export-ignore
-.travis.yml        export-ignore
-CODE_OF_CONDUCT.md export-ignore
-README.md          export-ignore
-docs               export-ignore
-tests              export-ignore
+* text=auto
+
+/.github                export-ignore
+/docs                   export-ignore
+/tests                  export-ignore
+/.editorconfig          export-ignore
+/.gitattributes         export-ignore
+/.gitignore             export-ignore
+/.kodiak.toml           export-ignore
+/CODE_OF_CONDUCT.md     export-ignore
+/ecs.php                export-ignore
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -1,7 +1,16 @@
...
+/tests                  export-ignore
+/utils                  export-ignore
+/.editorconfig          export-ignore
+/.gitattributes         export-ignore
+/.gitignore             export-ignore
+/.kodiak.toml           export-ignore
+/CODE_OF_CONDUCT.md     export-ignore
+/ecs.php                export-ignore
+/phpstan.neon           export-ignore
+/phpstan-extension.neon export-ignore
```
Kocal, line 14: To keep, otherwise we won't be able to use the PHPStan extension

.gitattributes
```diff
@@ -1,7 +1,16 @@
...
-CODE_OF_CONDUCT.md export-ignore
-README.md          export-ignore
-docs               export-ignore
-tests              export-ignore
+* text=auto
+
+/.github                export-ignore
+/docs                   export-ignore
+/tests                  export-ignore
+/utils                  export-ignore
```
Kocal, line 6: To keep, otherwise we won't be able to use the PHPStan extension

Kocal: Or you can be futur-proof and use the following I guess:
```gitignore
/utils/*
!/utils/phpstan-behaviors
```

TomasVotruba:
@ker0x Thank you. Could you process the feedback from @Kocal ?

ker0x:
@TomasVotruba It's done 😉

Kocal:
Thanks !)

TomasVotruba:
Thanks :+1:

PR #654, opened 2021-08-11
Static fixes, upgrade tests to PHP 7.4 and PHP 8.0 syntax

TomasVotruba:


src/Contract/Entity/TreeNodeInterface.php
```diff
@@ -37,6 +37,8 @@ public function setParentMaterializedPath(string $path): void;
 
     public function getParentNode(): ?self;
 
+    public function setParentNode(self $treeNode): void;
+
```
TomasVotruba, line 41: This method was missing in interface contract, but was part of provided methods trait. 
We hope the upgrade is ok. If you use custom methods, just add this one.

Issue #659, opened 2021-09-11
Use best practices for service registration


franmomu:
Following https://symfony.com/doc/current/bundles/best_practices.html#services:

> If the bundle defines services, they must be prefixed with the bundle alias instead of using fully qualified class names like you do in your project services. For example, AcmeBlogBundle services must be prefixed with acme_blog. The reason is that bundles shouldn’t rely on features such as service autowiring or autoconfiguration to not impose an overhead when compiling application services. 

> In addition, services not meant to be used by the application directly, should be defined as private. For public services, aliases should be created from the interface/class to the service id. For example, in MonologBundle, an alias is created from Psr\Log\LoggerInterface to logger so that the LoggerInterface type-hint can be used for autowiring.
> 
> Services should not use autowiring or autoconfiguration. Instead, all services should be defined explicitly.

It would be nice to follow these practices in:

https://github.com/KnpLabs/DoctrineBehaviors/blob/16cf5617882d4e5828215f2018cb58b3d24c0ee7/config/services.php#L17-L20

I could give it a try if this is fine, the only problem I see is that `UserProvider` is using:

https://github.com/KnpLabs/DoctrineBehaviors/blob/16cf5617882d4e5828215f2018cb58b3d24c0ee7/src/Provider/UserProvider.php#L12-L16

~`Security` which is registered by `symfony/security-bundle` which is not a dependency (BTW looks like `security.token_storage` could be used instead).~ I would leave the `Security` service as it is for now.

franmomu:
Closing here, see https://github.com/KnpLabs/DoctrineBehaviors/pull/660#issuecomment-917385552

PR #660, opened 2021-09-11
Register services explicitly

franmomu:
Since it wasn't too much trouble, I've done the first step of https://github.com/KnpLabs/DoctrineBehaviors/issues/659.

After this, it would be nice to :
- change these services to private
- change their names with the prefixed bundle 
- create public aliases with the FQCN 
- deprecate these aliases

### DIFF
--- a/config/services.php
+++ b/config/services.php
@@ -2,7 +2,23 @@
 
 declare(strict_types=1);
 
+use Knp\DoctrineBehaviors\Contract\Provider\LocaleProviderInterface;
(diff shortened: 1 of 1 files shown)


TomasVotruba:
The config you provided registers services manually one by one.

franmomu:
> The config you provided registers services manually one by one.

yep, that is what I understand (and what other bundles do) by "all services should be defined explicitly", to explicitly register services one by one with all the dependencies, tags, etc.

TomasVotruba:
That's the complexity we want to avoid.

franmomu:
ok then.

PR #664, opened 2021-10-01
fix isEmpty on zero string value

toucanbleu:
if $value is '0' then the translation is considered as empty


TomasVotruba:
Hi, thanks for the fix. 
We'll need a test case to make sure nobody will break it in the future.

PR #667, opened 2021-10-20
use ClassMetadataInfo to determine the referencedColumnName

jorenMartens:
In case you want to use something else then 'id'.

Is it also possible to make a 2.2.1 release for php 7 support if this is approved?


TomasVotruba:
Thank you :+1:

PR #673, opened 2021-11-10
fix generateSlugValue method signature

### LINKED ISSUE #672: generateSlugValue method has wrong return type
[This return type](https://github.com/KnpLabs/DoctrineBehaviors/blob/master/src/Model/Sluggable/SluggableMethodsTrait.php#L55) is wrong, it's probably left after a previous refactoring of the method.
The current code can't return anything but a string. The phpdoc should be removed and `: string` should be added instead.

garak:
Fix #672


TomasVotruba:
Thank you :+1:

PR #675, opened 2021-11-25
Fix Collection template usages

franmomu:
When using [Collection](https://github.com/doctrine/collections/blob/42a111afab917a328fdc38a35101c089ce52b580/lib/Doctrine/Common/Collections/Collection.php#L27-L32), it need to specify the type of the key.

In case of `Translatable` the key is the locale:

https://github.com/KnpLabs/DoctrineBehaviors/blob/a6a4bf3b118137de11b9913a76c58ae89f68c307/src/Model/Translatable/TranslatableMethodsTrait.php#L55-L57


TomasVotruba:
Thank you :+1:

Issue #676, opened 2021-11-30
LoggableTrait assumes numeric id


parijke:
The fuctions in the LoggableTrait assumes numeric id's, but with uuid's and uniqid() it is not certain.

TomasVotruba:
Resolved in https://github.com/KnpLabs/DoctrineBehaviors/pull/677

PR #677, opened 2021-11-30
Update LoggableTrait.php

parijke:
Use for non-numeric keys. Adresses issue #676


parijke:
@TomasVotruba how to handle getId() where it can be something else than a integer? I think you have created the PHPstan rule that fails.

TomasVotruba:
Thank you :+1: makes sense

TomasVotruba:
Thank you :+1: makes sense

PR #680, opened 2021-12-31
Update BlameableEventSubscriber.php

FabienPapet:
👋 

This PR fixes deprecation on doctrine bundle.

``` 
1x: Method "Doctrine\Common\EventSubscriber::getSubscribedEvents()" might add "array" as a native return type declaration in the future. Do the same in implementation "Knp\DoctrineBehaviors\EventSubscriber\BlameableEventSubscriber" now to avoid errors or add an explicit @return annotation to suppress this message.
``` 

Please let me know if I missed something, I'd be glad to help.


TomasVotruba:
Thank you :+1:

Issue #285, opened 2016-01-27
[Translatable] translatable_fetch_method: 'EAGER' still doesn't work


Hanmac:
I have a Vehicle class which does have a ManyToOne connection (lazy) to FuelType.
FuelType is a Translatable which is connected to FuelTypeTranslation (per translatable_fetch_method its 'EAGER')

but when i go to the side and wanted to output a Vehicle with its translated FuelType,
it still does do 2 queries for FuelType and one for FuelTypeTranslation.

Even if i do change the connection from Vehicle to FuelType into 'EAGER' too it still does make this two extra queries (even if it should has already the FuelType from the Vehicle Query), but it seems it doesn't work.

how can I do it that it does look for FuelType and its Translation in one Query? or is that a Doctrine bug?

waitxd:
#261

TomasVotruba:
I'm going through the old issues/PRs and closing most of them so we can focus on now.
This one is missing any feedback or reasoning so I cannot deduce anything from it.

Thanks for you though, but closing for reasons above.

If the bug still remains, please send failing test case to verify it. Only with test we'll be albe to help.

MichaelBrauner:
First - Thank you for your efforts and this/these great bundle(s)! I completly love it. 

But I have  the same a problem like the opener of this issue ... 

I translated some `Entity\Page` - everything works just fine. 
BUT when I query the pages the eager loading do not work at all. 

I have in my `services.yaml`:

```yaml
parameters:
  app_locales: de|en
  locale: en
  doctrine_behaviors_translatable_fetch_mode: "EAGER"
  doctrine_behaviors_translation_fetch_mode: "EAGER"
```

But it makes absolute no difference in my queries. 

```sql
SELECT
  t0.id AS id_1,
  t0.title AS title_2,
  t0.locale AS locale_3,
  t0.translatable_id AS translatable_id_4
FROM
  page_translation t0
WHERE
  t0.translatable_id = ?
```

This query is a distinct query every time no matter what.

TomasVotruba:
Thank you for appreciation :+1: 

I do not use this bundle much, so I don't have a clue. But I can make failing test pass :) could you share your use case in test case?

MichaelBrauner:
Yes. I can try that. Give me some days.

MichaelBrauner:
Ok. After digging - I see. 

[...]

PR #686, opened 2022-02-17
Update documentation for #285.

MichaelBrauner:
I created a piece of documentation for these lazy loading options. 
So it is very clear that no join on any table is getting performed.


TomasVotruba:
Thank you :+1:

PR #687, opened 2022-02-18
Translatable documentation update

Kerrialn:
- Convert translatable.md examples to use attributes
- Add example of how to use Translatable with API Platform in translatable-api-platform.md


TomasVotruba:
Thanks for the PR @Kerrialn and for the ping @MichaelBrauner 👍

Issue #307, opened 2016-03-18
Translatable and inheritance mapping


jacekpl:
Hi,

I've got two entities: TableOfContents and ContentType. Both of them have also Translatable entities.

```
<?php
namespace AppBundle\Entity;

use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\ORM\Mapping as ORM;
use Knp\DoctrineBehaviors\Model as ORMBehaviors;

/**
 * Class TableOfContents
 * @package AppBundle\Entity
 * @ORM\Entity()
 * @ORM\Table()
 * @ORM\InheritanceType("SINGLE_TABLE")
 * @ORM\DiscriminatorColumn(name="type", type="string")
 * @ORM\DiscriminatorMap({"tableOfContents" = "TableOfContents", "pdf" = "PdfType", "content" = "ContentType"})
 */
class TableOfContents
{
    use ORMBehaviors\Translatable\Translatable;

    /**
     * @var int
     * @ORM\Id()
     * @ORM\GeneratedValue(strategy="AUTO")
     * @ORM\Column(type="integer")
     */
    protected $id;

    public function getId()
    {
        return $this->id;
    }
}
```

```
<?php
namespace AppBundle\Entity;

use Doctrine\ORM\Mapping as ORM;
use Knp\DoctrineBehaviors\Model as ORMBehaviors;

/**
 * Class TableOfContentsTranslation
 * @package AppBundle\Entity
 * @ORM\Entity()
 */
class TableOfContentsTranslation
{
    use ORMBehaviors\Translatable\Translation;

    /**
     * @var string
     * @ORM\Column(type="string", length=255)
     */
    protected $title;

    public function getTitle()
    {
        return $this->title;
    }

    public function setTitle($title)
    {
        $this->title = $title;
    }
}
```

```
<?php
namespace AppBundle\Entity;

use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\ORM\Mapping as ORM;
use Knp\DoctrineBehaviors\Model as ORMBehaviors;

/**
 * Class ContentType
 * @package AppBundle\Entity
 * @ORM\Entity()
 * @ORM\HasLifecycleCallbacks()
 */
class ContentType extends TableOfContents
{
    use ORMBehaviors\Translatable\Translatable;

    const TYPE = "content";
    //(...)
}
```

```
<?php
namespace AppBundle\Entity;

use Doctrine\ORM\Mapping as ORM;
[...]

Issue #336, opened 2016-10-31
[Translatable] Joined Table translatable inheritance not working


paultarr:
Using knplabs/doctrine-behaviors": "1.4.0"

RE this fix (https://github.com/KnpLabs/DoctrineBehaviors/pull/317), this doesn't seem to work as expected. When generating the tables with "php app/console doctrine:schema:update --force", the base table translation fields are replicated in the subclass translation table, but without a translatable_id to work from.

Is there another inheritance mechanism that should be used here to make this work as expected i.e.

Base Table - Core non-translated fields
Base Translation Table - Core translated fields
Sub Class Table - Sub class non-translated fields
Sub Class Translation Table - Sub class translated fields

Example code:

Base Table (Purchasable)
~~~~
<?php

use Doctrine\ORM\Mapping as ORM;
use Sonata\MediaBundle\Provider\ImageProvider;
use JMS\Serializer\Annotation\ExclusionPolicy;
use JMS\Serializer\Annotation\Expose;
use Knp\DoctrineBehaviors\Model as ORMBehaviors;
use Sonata\TranslationBundle\Model\TranslatableInterface;


/**
 * @ORM\Entity
 * @ORM\InheritanceType("JOINED")
 * @ORM\DiscriminatorColumn(name="class_name", type="string")
 * @ExclusionPolicy("all")
 */
class Purchasable implements TranslatableInterface
{
    use ORMBehaviors\Translatable\Translatable;
    /**
     * @ORM\Id
     * @ORM\Column(type="integer")
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    protected $id;

     /**     
     * @ORM\Column(type="boolean")
     */
    protected $status = true;

    public function getId()
    {
        return $this->id;
    }

    public function setName($name)
    {
        $this->translate(null, false)->setName($name);
        return $this;
    }

    public function getName()
    {
        return $this->translate(null, false)->getName(); 
    }

    public function setStatus($status)
    {
        $this->status = $status;
        return $this;
    }

    public function getStatus()
    {
        return $this->status;
    }

    public function setLocale($locale)
    {
        $this->setCurrentLocale($locale);

        return $this;
    }

    public function getLocale()
    {
        return $this->getCurrentLocale();
    } 
}

~~~~
Base Table Translated (PurchasableTranslation)
~~~~
<?php
use Doctrine\ORM\Mapping as ORM;
use Knp\DoctrineBehaviors\Model as ORMBehaviors;

[...]

PR #688, opened 2022-02-19
Add doctrine join table inheritance compatibility

MichaelBrauner:
This PR adds a failing test and a fix for #307 #317 and #336.

**As a short summary:** 

Whenever you use doctrine joined inheritance you get a `Doctrine\DBAL\Schema\SchemaException: There is no column with name 'translatable_id' on table`. 

This is because a unique constraint is going to be set for the translation subclass but this class don't have this column. This comment describes it well and also suggest the fix: https://github.com/KnpLabs/DoctrineBehaviors/issues/307#issuecomment-231264716

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -105,6 +105,10 @@ parameters:
                 - tests/Fixtures/Entity/TranslatableCustomizedEntity.php
                 - tests/Fixtures/Entity/Translation/TranslatableCustomizedEntityTranslation.php
 
+        -
+            message: '#There should be no empty class#'
+            path: tests/Fixtures/Entity/Translatable/ExtendedTranslatableEntityWithJoinTableInheritanceTranslation.php
+
         # false positive, already checked for type
         -
             message: '#Cannot call method getName\(\) on ReflectionClass\|null#'
--- a/src/EventSubscriber/TranslatableEventSubscriber.php
+++ b/src/EventSubscriber/TranslatableEventSubscriber.php
@@ -28,9 +28,10 @@ final class TranslatableEventSubscriber implements EventSubscriberInterface
 
     public function __construct(
         private LocaleProviderInterface $localeProvider,
-        string $translatableFetchMode,
-        string $translationFetchMode
-    ) {
+        string                          $translatableFetchMode,
+        string                          $translationFetchMode
+    )
+    {
         $this->translatableFetchMode = $this->convertFetchString($translatableFetchMode);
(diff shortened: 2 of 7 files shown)


MichaelBrauner:
I will work on the failing checks.

TomasVotruba:
Hi, thanks for the PR :+1:  The coding standard can be fixed with:

```bash
composer fix-cs
```

See https://github.com/KnpLabs/DoctrineBehaviors#3-steps-to-contribute

MichaelBrauner:
A few PHPStan issues remain.

```shell
 ------------------------------------------------------------------------
  tests/Fixtures/Entity/Translatable/ExtendedTranslatableEntityWithJoinTableInheritanceTranslation.php:9
 ------------------------------------------------------------------------
  - '#There should be no empty class#'
  ```

This is a common usecase. Imagine you have a `Page` entity with a translated `title`. Than u want to have a `ContactPage` extended from `Page`. The translation has to extend PageTranslation but as long as you don't want to translate any `ContactPage`fields this class needs to be empty.

```shell
------------------------------------------------------------------------
  tests/Fixtures/Entity/Translatable/TranslatableEntityWithJoinTableInheritance.php:16
 ------------------------------------------------------------------------
  - '#Attribute must have all names explicitly defined#'
 ------------------------------------------------------------------------

 ------------------------------------------------------------------------
  tests/Fixtures/Entity/Translatable/TranslatableEntityWithJoinTableInheritanceTranslation.php:17
 ------------------------------------------------------------------------
  - '#Attribute must have all names explicitly defined#'
 ------------------------------------------------------------------------
```

This is the recommended implemenation. And the `#[Entity]` annotation is used this way inside the whole project.

Can I update the `phpstan.neon` file to ignore these messages?

TomasVotruba:
The ` - '#There should be no empty class#'` can be ignored.

The rest should be fixed to make argument name explicit.

TomasVotruba:
> And ... I don't undestand why the last Rector run failed. I did not change anything related to this.

Skip that for now, as it's probably related with new release. I'll check if after merge.

TomasVotruba:
Thank you :clap:

MichaelBrauner:
> Thank you 👏

Thank you for allowing me to do that. And for reacting so quickly.

I also will write some documentation related to this PR in the next few days.

TomasVotruba:
> I also will write some documentation related to this PR in the next few days.

That would be very useful improvement here. Thanks a lot for your work :+1:

Issue #690, opened 2022-03-19
getSingleIdentifierFieldName can cause a scheme error


connorhu:
This part can cause a bug if field name not equals with column name in your project:
https://github.com/KnpLabs/DoctrineBehaviors/blob/16a059fbc4d89b08cd0a46ca6ca0fd0f3fb7c9fd/src/EventSubscriber/TranslatableEventSubscriber.php#L129-L138

PR #691, opened 2022-03-19
fixed referencedColumnName #690

connorhu:
fixed referencedColumnName if fieldName != columnName #690


TomasVotruba:
Thanks for the fix 👍 
We'll need a test case too

krixer:
I have made a new updated PR (https://github.com/KnpLabs/DoctrineBehaviors/pull/744) to avoid conflicts with this one, which includes a new test for this case.

PR #695, opened 2022-04-25
Fix phpstan behaviors to allow an interface inherited from TranslatableInterface

nicolassing:
If I use my own TranslatableInterface instead of the one provided by the bundle, phpstan throw an error. 

Here an example 

```php
use Acme\Contract\TranslatableInterface;

class Product implements TranslatableInterface
{
}
```

```php
use Knp\DoctrineBehaviors\Contract\Entity\TranslatableInterface as KnpTranslatableInterface;

interface TranslatableInterface extends KnpTranslatableInterface
{
}
```

Thx !

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -105,6 +105,10 @@ parameters:
                 - tests/Fixtures/Entity/TranslatableCustomizedEntity.php
                 - tests/Fixtures/Entity/Translation/TranslatableCustomizedEntityTranslation.php
 
+        -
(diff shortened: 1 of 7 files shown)


nicolassing:
To be honest I don't know how to fix the Rector failed test :pleading_face:

TomasVotruba:
Try running the same command without `--dry-run`:

```bash
vendor/bin/rector process
```

PR #706, opened 2022-06-24
Fix phpstan behaviors to allow an interface inherited from TranslatableInterface

nicolassing:
If I use my own TranslatableInterface instead of the one provided by the bundle, phpstan throw an error. 

Here an example 

```php
use Acme\Contract\TranslatableInterface;

class Product implements TranslatableInterface
{
}
```

```php
use Knp\DoctrineBehaviors\Contract\Entity\TranslatableInterface as KnpTranslatableInterface;

interface TranslatableInterface extends KnpTranslatableInterface
{
}
```

Thx !

PS: Sorry I've closed the previous PR (#695) after a mistake :sweat:


TomasVotruba:
Thank you, great job :+1:

PR #710, opened 2022-09-06
Add generics annotations for translatable behavior

deguif:


src/Model/Translatable/TranslatablePropertiesTrait.php
```diff
@@ -5,18 +5,20 @@
...
+/**
+ * @template T of \Knp\DoctrineBehaviors\Contract\Entity\TranslationInterface
+ */
 trait TranslatablePropertiesTrait
 {
     /**
-     * @var Collection<string, TranslationInterface>
+     * @var Collection<string, T>
      */
     protected $translations;
```
TomasVotruba, line 17: This breaks the PHPStorm autocomplete. 
Could you share the gif with verification it won't happen?

Thanks

deguif: You're right, it's a little too early as template with restriction (`@template T of ..`) will be available in version 2022.3

src/Model/Translatable/TranslatableMethodsTrait.php
```diff
@@ -117,6 +128,9 @@ public function getDefaultLocale(): string
         return $this->defaultLocale;
     }
 
+    /**
+     * @return class-string<T>
+     */
     public static function getTranslationEntityClass(): string
     {
         return static::class . 'Translation';
```
deguif, line 136: PHPStan is complaining here as the class is not ensured to exist.
I see two alternatives (other than ignoring the errors) :
- Remove the `@return class-string<T>`.
- Add a check on the class name to check it exists (and throw an exception if it doesn't exist).

TomasVotruba: Which one would you prefer and why?

deguif: The `class-string<>` brings value here for static analysis as it can detect when class name is invalid / not existing.

I would be in favor to keep it, but it will require to add boilerplate on the default implementation code to check that the class exists before and maybe also check that's it's an implementation of `TranslationInterface`. When overriding the trait method, simply returning `class::name` will allow static analyzer to do its job and ensure the class is valid and an instance of TranslationInterface. WDYT?

TomasVotruba: To be honest... I don't know. The reason is I don't have time to maintain this package and give the propper care it needs.

It step up few years ago, when we used it at work and needed new PHP version support. Yet haven't used it since.

I have other OS project I want to focus on.

Saying that, I'll be stepping down as maintainer and give space someone to take over.

src/Model/Translatable/TranslatablePropertiesTrait.php
deguif: That is now working since PHPStorm introduced support for generics and this notation.

HeahDude:
Could we move forward on this? Can I help?

acrobat:
@HeahDude I think this should be tested with the latest phpstan 2 release, as this PR is open for a long time and phpstan has made huge steps since. 

We also need confirmation about the discussion around the `getTranslationEntityClass()` annotation -> https://github.com/KnpLabs/DoctrineBehaviors/pull/710#discussion_r963649666

I think otherwise this PR should be good to go and actually be quite the improvement!

PR #739, opened 2023-08-08
Fix github actions

raziel057:
Fix large part of the github actions. We still have an issue with rector because we use the dry-run parameter and there is some diff.

In that case, from the following commit (https://github.com/rectorphp/rector/pull/805), rector returns a Error: Process completed with exit code 2.

image

### DIFF
--- a/.github/workflows/code_checks.yaml
+++ b/.github/workflows/code_checks.yaml
@@ -32,7 +32,7 @@ jobs:
 
         steps:
             -   uses: actions/checkout@v2
-            -   uses: shivammathur/setup-php@v1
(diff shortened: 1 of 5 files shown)


raziel057:
@TomasVotruba @VincentLanglet Please tell me what is expected for the rector check. Do you wants to update the code according to the diff provided by the command: ``vendor/bin/rector process --dry-run --ansi`` or it's preferable to ignore it for now?

I can remove the dry-run parameter but I'm not sure that it's what you wants.

acrobat:
Closing in favor of #782

PR #741, opened 2023-09-08
Update TranslatableEventSubscriber.php

tacman:
getEntity is deprecated, use getObject()

https://github.com/doctrine/orm/issues/9875


acrobat:
Thanks for the PR @tacman, but this change is done through #782!

PR #738, opened 2023-08-08
Fix SF62 / SF63 deprecations (doctrine subscriber deprecated)

raziel057:
Fix deprecation raised when using ``doctrine/doctrine-bundle`` starting from v2.10 (see https://github.com/doctrine/DoctrineBundle/commit/656beec9a163b44987a7e79f89f74020d0358b17).

On my test suite (~400 tests) the deprecation is raised 7497x:
```
Remaining indirect deprecation notices (7497x)

833x: Since symfony/doctrine-bridge 6.3: Registering "Knp\DoctrineBehaviors\EventSubscriber\BlameableEventSubscriber" as a Doctrine subscriber is deprecated. Regisr it as a listener instead, using e.g. the #[AsDoctrineListener] attribute.
...
833x: Since symfony/doctrine-bridge 6.3: Registering "Knp\DoctrineBehaviors\EventSubscriber\LoggableEventSubscriber" as a Doctrine subscriber is deprecated. Regisr it as a listener instead, using e.g. the #[AsDoctrineListener] attribute.
...
833x: Since symfony/doctrine-bridge 6.3: Registering "Knp\DoctrineBehaviors\EventSubscriber\SluggableEventSubscriber" as a Doctrine subscriber is deprecated. Regisr it as a listener instead, using e.g. the #[AsDoctrineListener] attribute.
```

```
Since symfony/doctrine-bridge 6.3: Using Doctrine subscribers as services is deprecated, declare listeners instead
```

The test suite have been run locally after the change and I could fix :

image

Fix first part of https://github.com/KnpLabs/DoctrineBehaviors/issues/737

Issue #711, opened 2022-09-06
Looking for a maintainer


TomasVotruba:
Hello, everyone using this package :)

I don't have time to maintain this package and give it the proper care it needs. I stepped up a few years ago when we used it at work and needed new PHP version support. Yet haven't used it since.

I have another OS project I want to focus on, but I don't want to let this package be neglected.

Saying that I'll be stepping down as maintainer and giving space someone to take over, who is actively using this package and has the will to grow it :hugs:

tacman:
Although I'm not a good candidate for being the maintainer, I'm happy to be one of the maintainers, at least so that the tests run on my branch when I push.

I've made a few very minor changes and submitted them as PRs:  https://github.com/KnpLabs/DoctrineBehaviors/pulls

Kocal:
Hi everyone, I can try to take over this package if necessary.
I'm familiar with OSS and I've already submitted some patches to the package in the past. 

On the other hand, I only use the Translatable/SoftDeletable/Sluggable features.

TomasVotruba:
Hey, great to see willing folks here!

You need to reach out the knplabs, as I'm only collaborator here. 
That's how I got the position in the first place :) they might be slow to response, as their focus is out of PHP for years, so keep patience.

raziel057:
@TomasVotruba Good to see there is candidates to take care of this package. Would it be possible in the meantime to provide your inputs for the PR https://github.com/KnpLabs/DoctrineBehaviors/pull/739 so that I can finalize the fix of automatic checks / tests (Github Actions).

And if possible it would be nice also to accept the PR:
- https://github.com/KnpLabs/DoctrineBehaviors/pull/736 (to fix deprecation raised on PHP 8.2)
- https://github.com/KnpLabs/DoctrineBehaviors/pull/738 (to fix part of deprecation raised on SF6.3)

tacman:
@raziel057 With Symfony 7 just days away, are you still interested in being a maintainer? I'm happy to help as a collaborator.

tacman:
@TomasVotruba thanks for the work you did to revive this bundle.  I quite enjoyed the blog post about it, and even came back to this bundle for certain behaviors that weren't working for me in gedmo, in particular the translatable behavior.

I'm a bit nervous, though, because there are several outstanding PRs, and Symfony 7 is just days away.  The problem appears to be supporting things like Symfony 6.0 and 6.1, both unsupported by Symfony.  

[...]

PR #742, opened 2023-10-20
Update Subscribers with #[AsDoctrineListener]

DennisdeBest:
To remove the deprecation from implementing the EventSubscriberInterface this PR is to use the #[AsDoctrineListener] attribute instead.

To do so I set the `doctrine/doctrine-bundle` to `^2.7.2`.

Running the cs-fixer, phpstan and rector showed a couple of errors so I updated Rector and implemented some of the changes that those tools required.

There were errors in the github workflows so I updated the versions of the actions accordingly.

If there is more to update let me know I would be happy to help.

--- 

Looking at the current PRs this seems to be a combination of 

https://github.com/KnpLabs/DoctrineBehaviors/pull/739

https://github.com/KnpLabs/DoctrineBehaviors/pull/741

https://github.com/KnpLabs/DoctrineBehaviors/pull/738

https://github.com/KnpLabs/DoctrineBehaviors/pull/736


tfavareille (approved):

DennisdeBest:
I updated the tests some more and added support for Symfony 7.
I dropped support for Symfony5 and PHP 8.0

This kind of takes the PR [#750](https://github.com/KnpLabs/DoctrineBehaviors/pull/750) into account aswell.

azribilel (approved):

jgrygierek:
When it will be released?

Enz000:
Hello 👋 

Any updates about this PR?  😀

DennisdeBest:
> Hello 👋 
> 
> Any updates about this PR?  😀

Nothing so far 😕I use my fork on multiple projects waiting for it to get merged

jzecca:
Thanks a lot for the PR!
Would be great to have this released.

jgrygierek:
I really don't understand... What is the problem with releasing this one very important feature? It is just a few minutes...

virtualize:
@jgrygierek The problem is, that this library does not seem to have any active maintainers for quite some time. See #711

.github/workflows/code_checks.yaml
```diff
@@ -31,10 +31,10 @@ jobs:
         runs-on: ubuntu-latest
 
         steps:
-            -   uses: actions/checkout@v2
-            -   uses: shivammathur/setup-php@v1
+            -   uses: actions/checkout@v4
+            -   uses: shivammathur/setup-php@v2
                 with:
-                    php-version: 8.0
+                    php-version: 8.1
```
alexpozzi, line 37: This should be reverted as Symfony 6.0 still supports PHP 8.0.
If you want to add a test on PHP 8.1 you should add it instead of just change it to it IMHO.

.github/workflows/linter.yaml
```diff
@@ -11,11 +11,11 @@ jobs:
...
 
         steps:
-            -   uses: actions/checkout@v2
+            -   uses: actions/checkout@v4
 
-            -   uses: shivammathur/setup-php@v1
+            -   uses: shivammathur/setup-php@v2
                 with:
-                    php-version: 8.0
+                    php-version: 8.1
```
alexpozzi, line 18: This should be reverted as Symfony 6.0 still supports PHP 8.0.
If you want to add a test on PHP 8.1 you should add it instead of just change it to it IMHO.

.github/workflows/tests.yaml
```diff
@@ -7,16 +7,16 @@ jobs:
     tests:
         strategy:
             matrix:
-                php:
-                    - 8.0
+                operating-system: [ ubuntu-latest ]
+                php-versions: [ '8.1', '8.3' ]
```
alexpozzi, line 11: We should still test php 8.0 as well for the same reasons.

composer.json
```diff
@@ -11,18 +11,18 @@
         { "name": "Knplabs", "homepage": "http://knplabs.com" }
     ],
     "require": {
-        "php": ">=8.0",
+        "php": ">=8.1",
```
alexpozzi, line 14: This should be reverted as Symfony 6 still supports PHP 8.0 unless there is a reason for it.

composer.json
```diff
@@ -11,18 +11,18 @@
...
         "doctrine/dbal": "^3.3",
         "doctrine/orm": "^2.12",
-        "doctrine/doctrine-bundle": "^2.6",
-        "symfony/cache": "^5.4|^6.0",
-        "symfony/dependency-injection": "^5.4|^6.0",
-        "symfony/http-kernel": "^5.4|^6.0",
-        "symfony/security-core": "^5.4|^6.0",
-        "symfony/framework-bundle": "^5.4|^6.0",
-        "symfony/string": "^5.4|^6.0",
+        "doctrine/doctrine-bundle": "^2.7.2",
```
alexpozzi, line 19: Does this constraint need to be updated?

src/EventSubscriber/BlameableEventSubscriber.php
```diff
@@ -102,20 +108,20 @@ public function preUpdate(LifecycleEventArgs $lifecycleEventArgs): void
             return;
         }
 
-        $oldValue = $entity->getUpdatedBy();
-        $entity->setUpdatedBy($user);
+        $updatedBy = $object->getUpdatedBy();
```
alexpozzi, line 111:
```suggestion
        $oldUpdatedBy = $object->getUpdatedBy();
```

src/EventSubscriber/BlameableEventSubscriber.php
alexpozzi, line 1: I would keep `$entity` instead of changing it to `$object` as it's more specific to this case.

src/EventSubscriber/TranslatableEventSubscriber.php
alexpozzi, line 1: I would keep `$entity` instead of changing it to `$object` as it's more specific to this case.

src/EventSubscriber/UuidableEventSubscriber.php
alexpozzi, line 1: I would keep `$entity` instead of changing it to `$object` as it's more specific to this case.

tests/ORM/TreeNodeTest.php
alexpozzi, line 1: Why changing variable names?

tests/config/config_test.php
alexpozzi, line 1: I'm afraid that this change is not compatible to both Symfony 6 and Symfony 7.

alexpozzi (commented):
Thanks for your contribution!

I'm not super comfortable with this bundle as I don't use it myself, I mainly left you compatibility and maintainability comments.

In order to avoid conflicts, it would be better to avoid to change variables names (unless it has a real benefit in doing so).

composer.json
```diff
@@ -11,18 +11,18 @@
...
+        "symfony/cache": "^6.0|^7.0",
+        "symfony/dependency-injection": "^6.0|^7.0",
+        "symfony/http-kernel": "^6.0|^7.0",
+        "symfony/security-bundle": "^6.0|^7.0",
+        "symfony/framework-bundle": "^6.0|^7.0",
+        "symfony/string": "^6.0|^7.0",
```
endelwar: Why are versions 6.0-6.3, which are EOL, supported, while LTS version 5.4, which has support until November 2024 and security support until November 2025, is not?

composer.json
endelwar: The #AsDoctrineListener attribute was introduced in DoctrineBundle 2.7.2

composer.json
alexpozzi: It can be decided to introduce a breaking change in the library (like this one) and create a new major version while still supporting 5.4 (patches only) in the previous major version.
The fact that we bump the requirements here does not mean that the library stops supporting Symfony 5.4.

Supporting 6.0-6.3 makes sense as well unless there are some constraint that are creating us issues IMHO.
When writing libraries it's better to give a wider support of versions if there are no real reasons not to do it IMHO, not everyone might have upgraded yet to the latest 6.x version.

DennisdeBest:
@alexpozzi

Thanks for the reply.

I will take another look at it when I get back from vacation in two weeks.

.github/workflows/code_checks.yaml
connorhu: symfony 6.0 is "Unmaintained" https://symfony.com/releases/6.0
symfony 6.4 is maintained ATM and: "Requires: PHP 8.1.0 or higher"

jgrygierek:
Any news?

ernestoliberio:
the package is dead?

azribilel:
Hello @acrobat, Any news ?

tests/ORM/TreeNodeTest.php
DennisdeBest: this is done by Rector automatically : RenameVariableToMatchMethodCallReturnTypeRector

DennisdeBest:
There are some changes with Symfony 6.1 which make it difficult to use 6.0 as the lowest version. I updated the composer.json to only allow symfony 6.1 and php 8.1.

The changes  of some of the variables is the current Rector behaviour so we might aswell keep it like that. 

If this gets merged the next thing that might have to be updated is the doctrine ORM from v2 to V3 but doing that here makes it a bit too much in one PR.

MICMathieu:
@DennisdeBest some news about this PR ?

jgrygierek:
Anyone alive here?

jgrygierek:
Bump?

DennisdeBest:
@MICMathieu sorry no news so far

jgrygierek:
R.I.P DoctrineBehaviors?

jgrygierek:
Bump? xD

acrobat:
I'm trying to move the package forward and add compatibility to newer versions of Symfony and Doctrine. See https://github.com/KnpLabs/DoctrineBehaviors/issues/743#issuecomment-3315270632 and #782. So please test these changes in your project and provide feedback if you have the time. Thanks!

acrobat:
Thanks for your work @DennisdeBest! This is now implemented through https://github.com/KnpLabs/DoctrineBehaviors/pull/782

